PR #13, opened 2023-03-21
Separate core NAM code from other plugin dsp code

mikeoliphant:
This PR separates the core NAM code from other plugin dsp code and removes IPlug dependencies for NAM code.

I tried to keep this as simple and targeted as possible. It looks like bigger changes than it is because of the separation of dsp.h/dsp.cpp files into an IPlug dsp base class for noise gate, IR and EQ and the core dsp code for the network models.

I kept the input to the model processing as "double" for now - since that is what the IPlug plugin is using. I also haven't made any changes to the redundant gain handling - I'll save that for a separate PR.

With the changes in this PR, it is now easy to use the core NAM code directly in a context other than the IPlug plugin.


mikeoliphant:
This will also require some changes to the plugin project structures to link to the new source code locations.

Are the various VS project files generated somehow, or is that done manually on a per-project basis?

mikeoliphant:
@sdatkinson - I know you've got a lot on your plate, but maybe now is a good time to think about merging this, since you just released a new version of the plugin?

I've got some performance enhancements I'd like to make, but I don't want to make any more changes until this bit gets resolved.

dsp/dsp.h
```diff
@@ -10,296 +8,6 @@
 #include <vector>
 
 #include "IPlugConstants.h"
-#include <Eigen/Dense>
```
sdatkinson, line 13: My understanding is that this is a straight move over to `NAM/dsp.h`. Is that correct? Let me know if there are any other changes.

dsp/dsp.cpp
```diff
@@ -8,494 +8,6 @@
 #include <unordered_set>
 
 #include "dsp.h"
-#include "json.hpp"
```
sdatkinson, line 11: This is moved to `NAM/dsp.cpp` with no modification? Please confirm.

sdatkinson (commented):
Thanks for your patience waiting for this.

Quick confirmation on the "move" Qs and we'll be good!

dsp/dsp.h
mikeoliphant: It moves over the NAM-specific bits of dsp.h to NAM/dsp.h. So the NAM-specific stuff then gets removed from dsp/dsp.h - it just has the base class for the noise gate, IR and EQ  stuff.

mikeoliphant: There is only one other change. The NAM/dsp.h code removes the IPlug dependency - which just amounts to substituting "double" for "iplug::sample"

dsp/dsp.cpp
mikeoliphant: Like dsp.h - dsp.cpp gets split between dsp/dsp.cpp and NAM/dsp.cpp. The former has the noise gate, IR and EQ base class, and the latter has the network model stuff.

It also removes IPlug dependencies by using "double" instead of "iplug::sample".

mikeoliphant:
Let me know if anything is still unclear. It looks like a lot of stuff is going on, but it is basically just moving the NAM code into a separate folder from the other dsp code - slightly complicated by the fact that code was shared in dsp.h/dsp.cpp so  those files had to be split.

mikeoliphant:
Since this PR now has conflicts, I could maybe resubmit - perhaps one simpler PR to separate dsp.h/dsp.cpp (perhaps calling the NAM versions NAM.h/Nam.cpp), and a subsequent one for the removal of the IPlug dependency, and then one for moving files to a separate folder?

Just let me know if that would make things easier than handling this PR as is.

sdatkinson:
> Since this PR now has conflicts, I could maybe resubmit - perhaps one simpler PR to separate dsp.h/dsp.cpp (perhaps calling the NAM versions NAM.h/Nam.cpp), and a subsequent one for the removal of the IPlug dependency, and then one for moving files to a separate folder?
> 
> Just let me know if that would make things easier than handling this PR as is.

I'll have a look tonight and let you know. Thanks for your patience :)

sdatkinson:
@mikeoliphant [`issue-13`](https://github.com/sdatkinson/NeuralAmpModelerCore/tree/issue-13) resolves merge conflicts. Should be able to merge it into your branch and the PR should be good to go!

sdatkinson:
Got it in #21 :)

PR #15, opened 2023-04-01
Remove iplug2 dependency

olilarkin:
Possible duplicate of #13, which i only just spotted

This doesn't make changes w.r.t to Eigen or json.hpp, which should be dependencies of this repo IMO

also renames format.sh and uses pragma once for all include guards

### DIFF
--- a/dsp/ImpulseResponse.cpp
+++ b/dsp/ImpulseResponse.cpp
@@ -10,23 +10,22 @@
 
 #include "ImpulseResponse.h"
 
-dsp::ImpulseResponse::ImpulseResponse(const WDL_String &fileName,
+dsp::ImpulseResponse::ImpulseResponse(const char* fileName,
                                       const double sampleRate)
     : mWavState(dsp::wav::LoadReturnCode::ERROR_OTHER) {
   // Try to load the WAV
-  this->mWavState =
-      dsp::wav::Load(fileName, this->mRawAudio, this->mRawAudioSampleRate);
+  this->mWavState = dsp::wav::Load(fileName, this->mRawAudio,
+                                   this->mRawAudioSampleRate);
   if (this->mWavState != dsp::wav::LoadReturnCode::SUCCESS) {
     std::stringstream ss;
-    ss << "Failed to load IR at " << fileName.Get() << std::endl;
+    ss << "Failed to load IR at " << fileName << std::endl;
   } else
     // Set the weights based on the raw audio.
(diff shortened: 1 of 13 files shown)


mikeoliphant:
Yes - PR https://github.com/sdatkinson/NeuralAmpModelerCore/pull/13 removes the IPlug2 dependency, and separates the NAM model code from the other dsp code.

I didn't worry about removing the IPlug dependency from the other dsp code (noise gate, IR, EQ) - since they have other "IPluggy" structure to them. To be honest, I think that code could live in the plugin repo rather than here.

I went with "double" instead of "float", since that was what the plugin is currently using. I think maybe using a template class so as to be able to accept either double or float I/O might make sense, but I wanted to get the separation done first.

olilarkin:
Agree, ideally this repo would just be the model parsing and NN inference code. @sdatkinson iPlug already has EQ code: https://github.com/iPlug2/iPlug2/blob/master/IPlug/Extras/SVF.h which might be nicer. Also there are nice audio file-loading and convolution libraries out there!

sdatkinson:
Getting to this now. Resolving conflicts and jotting down notes:

* This PR keeps NAM DSP code in `dsp/dsp.[cpp,h]`. main moves to `NAM/`.  I think `NAM` is fine.
* Use `float` instead of `double` in place of `iplug::sample`. `double` is preferred for now, can let the models override for their own preference. I'd presume that DSP in single-precision is uncommon since "classical" numerical methods are probably the default, so I'm fine to let the models explicitly indicate how they're breaking from the norm at least for now.
* There was an `#include "IPlugConstants.h"` in `dsp/dsp.h` that I'm not sure what it was doing...

Unfortunately this series of PRs is going to break [NeuralAmpModelerPlugin](https://github.com/sdatkinson/NeuralAmpModelerPlugin) so it's going to be a moment before I can figure out whether things are all good or if there were bugs introduced in this (I know, I know). Just hang tight. I'll try to catch it up and iron out things if needed.

But I've got the merge conflicts resolved; I'm going to open a separate PR that appends the resolution onto your commit, @olilarkin and merge that.

Thanks all for your help--I really appreciate it!

mikeoliphant:
> * There was an `#include "IPlugConstants.h"` in `dsp/dsp.h` that I'm not sure what it was doing...

It was just required to use "iplug:sample" - so no longer required if those have been removed.

sdatkinson:
Shipped via #22

PR #24, opened 2023-04-02
fabs->fabsf in fast tanh

mikeoliphant:
This makes sure we don't get double promotion in the fast tanh function.

I suspect that fast tanh will be superseded by hard tanh before it sees the light of day, but just in case...


sdatkinson (approved):
LGTM! 👍🏻

PR #26, opened 2023-04-02
Separate out convnet

mikeoliphant:
The separates out the convnet code from dsp.h/dsp.cpp to be consistent with the other models.

Not really a big deal, but cleaner this way and I figure it makes sense to do now while we're moving stuff around.


sdatkinson (approved):
LGTM! 👍🏻

PR #28, opened 2023-04-03
Activation function refactor

mikeoliphant:
This PR refactors the activation function handling. It does the following:

- centralizes activation functions in activation.h/activation.cpp
- does function lookup by name at model initialization
- ensures that all activation function application is done in efficient block operations

It also adds a flag to turn on and off the fast tanh approximation. You just need to do call "**activations::Activation::enable_fast_tanh()**" before loading a model. Even with the potential of "Hardtanh" replacing it, it probably makes sense to implement this given all of the existing models out there that could be sped up. Maybe a toggle switch in the UI ("Fast Mode")?

I haven't yet switched over the LSTM code, and there are also some hardcoded sigmoids in the WaveNet code.

### DIFF
--- a/NAM/activations.cpp
+++ b/NAM/activations.cpp
@@ -1,7 +1,7 @@
 #include "activations.h"
 
 std::unordered_map<std::string, activations::Activation*> activations::Activation::_activations = {
-  {"Tanh", new activations::ActivationHardTanh()},
+  {"Tanh", new activations::ActivationTanh()},
   {"Hardtanh", new activations::ActivationHardTanh()},
   {"Fasttanh", new activations::ActivationFastTanh()},
   {"ReLU", new activations::ActivationReLU()},
--- a/NAM/activations.cpp
+++ b/NAM/activations.cpp
@@ -1,5 +1,3 @@
-#pragma once
-
 #include "activations.h"
 
 std::unordered_map<std::string, activations::Activation*> activations::Activation::_activations = {
--- a/NAM/activations.cpp
+++ b/NAM/activations.cpp
@@ -1,0 +1,10 @@
+#pragma once
+
+#include "activations.h"
+
+std::unordered_map<std::string, activations::Activation*> activations::Activation::_activations = {
+  {"Tanh", new activations::ActivationHardTanh()},
+  {"Hardtanh", new activations::ActivationHardTanh()},
+  {"Fasttanh", new activations::ActivationFastTanh()},
+  {"ReLU", new activations::ActivationReLU()},
+  {"Sigmoid", new activations::ActivationSigmoid()}};
--- a/NAM/activations.cpp
+++ b/NAM/activations.cpp
@@ -6,3 +6,30 @@ std::unordered_map<std::string, activations::Activation*> activations::Activatio
   {"Fasttanh", new activations::ActivationFastTanh()},
   {"ReLU", new activations::ActivationReLU()},
   {"Sigmoid", new activations::ActivationSigmoid()}};
+
+activations::Activation* tanh_bak = nullptr;
+
+activations::Activation* activations::Activation::get_activation(const std::string name)
+{
+  if (_activations.find(name) == _activations.end())
+    return nullptr;
+
+  return _activations[name];
+}
+
+void activations::Activation::enable_fast_tanh()
+{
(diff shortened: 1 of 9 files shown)


mikeoliphant:
Ok - I think there are more improvements that could be made here, but I'm going to leave it for now.

These changes shouldn't modify any existing behavior (other than making ConvNet a bit faster), or change the interface with the plugin (other than adding the enable/disable fast_tanh).

NAM/activations.cpp
```diff
@@ -0,0 +1,35 @@
+#include "activations.h"
+
+std::unordered_map<std::string, activations::Activation*> activations::Activation::_activations = {
+  {"Tanh", new activations::ActivationTanh()},
+  {"Hardtanh", new activations::ActivationHardTanh()},
+  {"Fasttanh", new activations::ActivationFastTanh()},
+  {"ReLU", new activations::ActivationReLU()},
+  {"Sigmoid", new activations::ActivationSigmoid()}};
```
jatinchowdhury18, line 8: Seems like this memory will be leaked?

NAM/activations.h
```diff
@@ -1,15 +1,125 @@
...
+  }
+  virtual void apply(Eigen::Block<Eigen::MatrixXf> block)
+  {
+    apply(block.data(), block.rows() * block.cols());
+  }
+  virtual void apply(Eigen::Block<Eigen::MatrixXf, -1, -1, true> block)
+  {
+    apply(block.data(), block.rows() * block.cols());
+  }
+  virtual void apply(float* data, long size) {}
```
jatinchowdhury18, line 51: Putting functions which will be called per-sample (or maybe multiple times per-sample depending on the network architecture) behind a virtual interface will likely have a significant performance impact, since it prevents the compiler from inlining these function calls in most cases.

In [RTNeural](https://github.com/jatinchowdhury18/RTNeural), this is a big reason why the "dynamic" API is much slower than the "static" API.

NAM/activations.h
```diff
@@ -1,15 +1,125 @@
...
+class ActivationTanh : public Activation
 {
-  return 1.0f / (1.0f + expf(-x));
+  public:
+    void apply(float *data, long size) override
+    {
+      for (long pos = 0; pos < size; pos++)
+      {
+        data[pos] = std::tanh(data[pos]);
+      }
```
jatinchowdhury18, line 69: It would probably be better to use Eigen's math functions for implementing some of these things, since they do some internal vectorization. (And that vectorization can be even better if the Matrix size is known at compile-time)

NAM/activations.cpp
mikeoliphant: Yeah, it makes me uncomfortable, too - even though it is global, so should happen once and stick around for the lifetime of the program. My c++ chops are rusty - not sure what the preferred modern c++ way of handling this is.

I also thought about creating individual global instances and just taking a pointer to them in the map.

NAM/activations.h
mikeoliphant: I was worried about inlining, too, which is why I made sure to keep everything as block operations. I did not see any slowdown over the existing implementation in my testing, but that could be compiler-specific? I'm testing using VS on Windows.

The override methods definitely make it more convenient to use, but it would be worth giving that up if there is a clearly demonstrable performance gain to be had.

NAM/activations.h
mikeoliphant: To be honest, tanh is best avoided completely...

I'm pretty sure I looked for a way to do a matrix-level tanh with Eigen, and didn't see one.

mikeoliphant:
Btw, @jatinchowdhury18 - your work on rtneural (among other things) is much appreciated. I haven't yet used it directly myself, but is clearly has been very helpful throughout the audio ML community.

I considered trying to port NAM to sit on top of rtneural, but decided to focus on smaller changes instead.

sdatkinson (approved):
If you wanna grab the memory leak w/ what I wrote then it seems perfect to me. If not, I'll still merge & fix myself.

Thanks!

mikeoliphant:
> If you wanna grab the memory leak w/ what I wrote then it seems perfect to me. If not, I'll still merge & fix myself.

What you wrote is pretty much what I was thinking, so how about you go ahead and merge and add it.

PR #31, opened 2023-04-11
added dspData struct and appropriate get_dsp() -methods to work with it.

masqutti:
Let's try this again. :) Just the dspData struct addition, I added the metadata as a JSON as I don't know the exact properties (and I guess they are still on poll anyways?) but we can always add more specific data in in the future I guess.

### DIFF
--- a/NAM/dsp.h
+++ b/NAM/dsp.h
@@ -10,6 +10,7 @@
 #include <Eigen/Dense>
 
 #include "activations.h"
+#include "json.hpp"
 
 enum EArchitectures
 {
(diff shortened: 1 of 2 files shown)


sdatkinson (changes requested):
I really like this!

However, there's some typing that isn't quite correct, and there's the addition of the new filesystem implementation that I don't think will compile as-is?

Since those might cause problems, it would probably be easiest to decrease the scope of this PR to defining `dspData` and extending the `get_dsp()` interface.

sdatkinson:
Resolves #45.

masqutti:
Sigh! Somehow I thought that new pushes on this branch would not propagate to the pull request automatically... Little did I know. I will, once again, create a new pull request with just the dspStruct addition 🙃

PR #32, opened 2023-04-13
Activation constructor/destructor cleanup

mikeoliphant:
This PR just cleans up the Activation class constructor/destructors. It wasn't currently a problem (although it could have been in the future), but it gits rid of a bunch of annoying compiler warnings.


sdatkinson (approved):
LGTM! 👍🏻

PR #33, opened 2023-04-14
Method to determine whether model has loudness data

pawelKapl:
Adding method determining whether model has configured loudness parameter. If current loudness is equal to the default one - normalisation wont occur. This is a compromise to avoid passing unnecessary bool param to the model.

(partially) Fixes sdatkinson/NeuralAmpModelerPlugin#183

This is a first of 2 parts to resolve ☝️ issue.

### DIFF
--- a/NAM/dsp.h
+++ b/NAM/dsp.h
@@ -60,6 +60,7 @@ public:
   //   DSP subclass implementation.
   virtual void finalize_(const int num_frames);
   void SetNormalize(const bool normalize) { this->mNormalizeOutputLoudness = normalize; };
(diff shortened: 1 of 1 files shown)


NAM/dsp.h
```diff
@@ -60,6 +60,7 @@ class DSP
   //   DSP subclass implementation.
   virtual void finalize_(const int num_frames);
   void SetNormalize(const bool normalize) { this->mNormalizeOutputLoudness = normalize; };
+  bool HasLoudness() { return mLoudness != TARGET_DSP_LOUDNESS; };
```
sdatkinson, line 63: Eh, that'll work for now.

sdatkinson (approved):
LGTM! 👍🏻

PR #34, opened 2023-04-17
Add CMake build tools

mikeoliphant:
The PR adds  a CMake build processes to allow checking buildability as well as creating some tools.

Right now, the only tool is a simple "loadmodel" program, but it is sufficient to test whether the core NAM code compiles.

To enable building, I added the Eigen and json.hpp dependencies to this repo.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,24 @@
+name: Build
+
+on:
+  [workflow_dispatch, pull_request]
+
+env:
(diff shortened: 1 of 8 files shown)


mikeoliphant:
I added a simple model benchmarking tool.

It would be great to get this merged - it makes it much easier to experiment with the core code, and gives us build checks on PRs.

sdatkinson (approved):
Seems good to me.

Thanks for this; excited to get into it more!

PR #42, opened 2023-05-06
Add wav error message string function to dsp::wav

olilarkin:
Whilst working on the plug-in code it seemed moving this here would make sense


sdatkinson (approved):
LGTM! 👍🏻

PR #43, opened 2023-05-08
Add fast tanh and fast sigmoid to LSTM

mikeoliphant:
This PR adds fast tanh and fast sigmoid activation for LSTM models. It keys off of the enable_fast_tanh()/disable_fast_tanh() methods.

My testing indicates around a 40% performance improvement for LSTM models.

### DIFF
--- a/NAM/activations.cpp
+++ b/NAM/activations.cpp
@@ -6,6 +6,8 @@ activations::ActivationHardTanh _HARD_TANH = activations::ActivationHardTanh();
 activations::ActivationReLU _RELU = activations::ActivationReLU();
 activations::ActivationSigmoid _SIGMOID = activations::ActivationSigmoid();
 
+bool activations::Activation::using_fast_tanh = false;
+
 std::unordered_map<std::string, activations::Activation*> activations::Activation::_activations =
   {{"Tanh", &_TANH}, {"Hardtanh", &_HARD_TANH}, {"Fasttanh", &_FAST_TANH}, {"ReLU", &_RELU}, {"Sigmoid", &_SIGMOID}};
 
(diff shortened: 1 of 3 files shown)


NAM/activations.h
```diff
@@ -32,6 +32,12 @@ inline float fast_tanh(const float x)
           / (2.44506634652299f + (2.44506634652299f + x2) * fabsf(x + 0.814642734961073f * x * ax)));
 }
 
+inline float fast_sigmoid(const float x)
+{
+  return 0.5f * (fast_tanh(x * 0.5f) + 1.0f);
```
sdatkinson, line 37: Looks correct ✅

NAM/activations.cpp
```diff
@@ -6,6 +6,8 @@ activations::ActivationHardTanh _HARD_TANH = activations::ActivationHardTanh();
 activations::ActivationReLU _RELU = activations::ActivationReLU();
 activations::ActivationSigmoid _SIGMOID = activations::ActivationSigmoid();
 
+bool activations::Activation::using_fast_tanh = false;
```
sdatkinson, line 9: I guess this is a separate but non-conflicting way to do this, then? FWIW, I think I'm happy to have it be global like this 👍🏻

sdatkinson (approved):
LGTM! 👍🏻

NAM/activations.cpp
mikeoliphant: > I guess this is a separate but non-conflicting way to do this, then?

Because the LSTM code calls the activation functions directly (since it is operating on individual weights), it needed a flag to key off of.

PR #46, opened 2023-05-25
dspStruct pull request

### LINKED ISSUE #45: Extend `get_dsp()` interface
Extend `get_dsp()` to accept not only accept file paths but model data.  This requires defining a `struct` to contain the information that is usually loaded in from a model file.

masqutti:
Yarr, we meet again!

I've now branched this dspStruct thingy into a separate branch, updated to latest changes and let's hope it works this time. :o) I've compiled without errors, NAM runs and only the dspStruct addition is now present in this pull request.

Original text from #31 : Just the dspData struct addition, I added the metadata as a JSON as I don't know the exact properties (and I guess they are still on poll anyways?) but we can always add more specific data in in the future I guess.

This replaces #31 and resolves #45

### DIFF
--- a/NAM/dsp.h
+++ b/NAM/dsp.h
@@ -10,6 +10,7 @@
 #include <Eigen/Dense>
 
 #include "activations.h"
+#include "json.hpp"
 
 enum EArchitectures
 {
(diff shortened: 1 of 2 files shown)


sdatkinson (approved):
Yes. Rock & roll! 🎸

PR #48, opened 2023-05-29
Allow NAM_SAMPLE_FLOAT to switch model input to float instead of double

mikeoliphant:
This PR adds a NAM_SAMPLE #define that is used instead of "double" for input samples for model processing.

If NAM_SAMPLE_FLOAT is defined, this will switch NAM_SAMPLE to "float". Otherwise it remains "double", so no there is no change in existing behavior.

This allows processing loops using "float" to avoid having to convert buffers to "double".

### DIFF
--- a/NAM/convnet.cpp
+++ b/NAM/convnet.cpp
@@ -99,7 +99,7 @@ convnet::ConvNet::ConvNet(const int channels, const std::vector<int>& dilations,
 {
 }
 
-convnet::ConvNet::ConvNet(const double loudness, const int channels, const std::vector<int>& dilations,
+convnet::ConvNet::ConvNet(const NAM_SAMPLE loudness, const int channels, const std::vector<int>& dilations,
                           const bool batchnorm, const std::string activation, std::vector<float>& params)
 : Buffer(loudness, *std::max_element(dilations.begin(), dilations.end()))
 {
--- a/NAM/dsp.cpp
+++ b/NAM/dsp.cpp
@@ -23,15 +23,15 @@ DSP::DSP()
 {
 }
 
-DSP::DSP(const double loudness)
+DSP::DSP(const NAM_SAMPLE loudness)
 : mLoudness(loudness)
 , mNormalizeOutputLoudness(false)
 , _stale_params(true)
 {
 }
 
-void DSP::process(double** inputs, double** outputs, const int num_channels, const int num_frames,
-                  const double input_gain, const double output_gain,
+void DSP::process(NAM_SAMPLE** inputs, NAM_SAMPLE** outputs, const int num_channels, const int num_frames,
+                  const NAM_SAMPLE input_gain, const NAM_SAMPLE output_gain,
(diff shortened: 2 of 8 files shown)


daleonov:
Good stuff. JUCE, for instance, uses float buffers, so keeping an extra double-precision buffer as an adapter for NAM is a bit annoying.

NAM/convnet.cpp
```diff
@@ -99,7 +99,7 @@ convnet::ConvNet::ConvNet(const int channels, const std::vector<int>& dilations,
 {
 }
 
-convnet::ConvNet::ConvNet(const double loudness, const int channels, const std::vector<int>& dilations,
+convnet::ConvNet::ConvNet(const NAM_SAMPLE loudness, const int channels, const std::vector<int>& dilations,
```
sdatkinson, line 102: Nit: I don't like `loudness` being of "sample" type--it's a parameter of the model.

But I get that it mainly serves to do arithmetic _with_ samples, so I can live with it.

NAM/dsp.cpp
```diff
@@ -23,15 +23,15 @@ DSP::DSP()
...
 : mLoudness(loudness)
 , mNormalizeOutputLoudness(false)
 , _stale_params(true)
 {
 }
 
-void DSP::process(double** inputs, double** outputs, const int num_channels, const int num_frames,
-                  const double input_gain, const double output_gain,
+void DSP::process(NAM_SAMPLE** inputs, NAM_SAMPLE** outputs, const int num_channels, const int num_frames,
+                  const NAM_SAMPLE input_gain, const NAM_SAMPLE output_gain,
```
sdatkinson, line 34: Yeah, it looks weird having all of these parameters being called "samples". Can you point me to somewhere else that follows this convention? It just doesn't sound right.

sdatkinson (approved):
Approving since I've called it a nit, but I'd appreciate hearing back from you about having gain parameters be of `NAM_SAMPLE` type. It feels odd, but I know I haven't seen enough DSP code to feel confident about conventions in the space.

NAM/dsp.cpp
mikeoliphant: I wasn't sure either, and I'm fine either way. "loudness" is an internal parameter anyway, and "input_gain" and "output_gain" should be probably be removed since the plugin just passes "1.0" and handles gain itself (as does my LV2 plugin).

daleonov:
By the way, what about ```Eigen::MatrixXf``` and ```Eigen::VectorXf```? Shouldn't they match buffer precision instead of always being float?

mikeoliphant:
> By the way, what about `Eigen::MatrixXf` and `Eigen::VectorXf`? Shouldn't they match buffer precision instead of always being float?

No - the models operate with floats regardless of the input sample precision.

mikeoliphant:
Ok, I've reverted the parameters to double.

sdatkinson (approved):
LGTM! 👍🏻

PR #50, opened 2023-06-09
Fix minor incompatibilities with clang/g++ build on Linux

daleonov:
Some errors came up when I was building a JUCE project in Linux. So, quick fix.


sdatkinson (approved):
LGTM! 👍🏻

PR #53, opened 2023-06-15
Fix compiler warnings

falkTX:
This PR fixes several compiler warnings generated by GCC under Linux.
They are quite a few but related to just 2 issues:
1. Comparison of signed vs unsigned types (because `std::vector<>::size()` returns `size_t` which is unsigned)
2. C++ class constructors have wrong member initialization order


sdatkinson (approved):
LGTM! 👍🏻

PR #63, opened 2023-07-09
Fix some wav impulse files would sometimes fail to load ... 

fab672000:
when the subchunk1Size>16 because the rounded to dword offset calculation was wrong.
Now IR's such as BD_HV_Creamback3_mixed.wav  (in [free pack from Bogren Digital Impulses](https://bogrendigital.com/pages/free-ir-pack)) load perfectly while other common cases (chunk size = 16) still work fine.


sdatkinson (approved):
LGTM! 👍🏻

PR #68, opened 2023-07-28
Allow defining DSP_SAMPLE_FLOAT to have dsp classes use float I/O

mikeoliphant:
This replaces double I/O for the dsp classes with DSP_SAMPLE. It is set to double by default (so no change for the NAM plugin) but can be switched to float by setting DSP_SAMPLE_FLOAT.

This is similar to the change made to the NAM code. I used a separate define to allow the dsp code to run at a different sample rate if needed/desired.

I just changed the I/O and left parameters as double.


sdatkinson (approved):
LGTM! 👍🏻

Issue #77, opened 2023-10-02
Make core NAM interfaces mono


mikeoliphant:
Along with the other cleanup of the core interfaces, I think it makes sense to consider making the model processing mono (rather than multi-channel).

There isn't any internal support for more than a single channel, and it would make the interface cleaner.

sdatkinson:
Great suggestion. I like it.

mikeoliphant:
It also has the benefit of making https://github.com/sdatkinson/NeuralAmpModelerCore/issues/27 much simpler to resolve.

mikeoliphant:
Btw, I'd be happy to do a PR for this and https://github.com/sdatkinson/NeuralAmpModelerCore/issues/27, along with removing the input/output gain. No worries if you'd rather do it yourself, though.

sdatkinson:
Go for it! 🙂

Issue #27, opened 2023-04-02
Reduce input/output buffer copies in model processing


mikeoliphant:
Currently, there are at least two buffer copies (one at input and one at output) that could be avoided by directly using the passed in input/output buffers.

olilarkin:
I agree. The whole allocating arrays of raw input/output pointers stuff in both NeuralAmpModelerCore and the Plug-in code is very suspicious and probably the source of the memory leaks

Issue #49, opened 2023-06-05
Do not allocate memory during WaveNet processing


mikeoliphant:
Currently, the WaveNet model processing code re-sizes vectors and matrices based on the audio buffer size during processing. This is non-ideal for real-time operation. Instead, all sizing operations should be done out-of-band of the processing loop.

In most cases, the current behavior should not cause significant problems. If there is a fixed audio buffer size the resize operations should only happen once. A fixed buffer size is not guaranteed, however - DAWs will sometimes vary the block size.

daleonov:
Agreed. There should be some kind of ```prepareBuffers()``` method. There's always some method that called by the DAW every time maximum block size or sample rate changes (```prepareToPlay()``` in JUCE, and there was something similar in iPlug), so at least happens once when the session starts, and it happens outside audio thread. It's the only safe place to resize the buffers.
Also, like Mike mentioned, block size can vary. Good example is when you get looped selection in the DAW, the last block in that loop is always smaller than usual, and then the next one is back to normal size. The trick is there's usually no ```prepareToPlay()``` callback from the DAW in that case, so you have to size down, and then size back up with no memory allocation.

mikeoliphant:
Yes - typically this would be handled by allocating for the max size, but then only processing the number of samples you are given.

Eigen has a way to specify the maximum matrix size, but unfortunately it is at compile time:

https://eigen.tuxfamily.org/dox/classEigen_1_1Matrix.html

It should be possible to create matrices/vectors at max size, and then just do block operations on them at the current given size.

sdatkinson:
Sounds reasonable--I had started to move some things out in the iPlug2 plugin, but yeah, this reeks of me getting back into C++ for this project 😅 

> Good example is when you get looped selection in the DAW, the last block in that loop is always smaller than usual, and then the next one is back to normal size.

Did not know--great call!

Let me know if either of you want to take this on--happy to assign it 👍🏻

mikeoliphant:
> yeah, this reeks of me getting back into C++

I feel your pain...

> Let me know if either of you want to take this on--happy to assign it

I can probably look into sorting this out.

olilarkin:
I just ran NAM audiounit with Apple's auval real time safety checker and it pointed to a couple of things... ```

Realtime-safety violation:
                libsystem_malloc.dylib`malloc
                NeuralAmpModeler`Eigen::internal::aligned_malloc(unsigned long)+0x38
[...]

PR #78, opened 2023-10-02
Processing interface cleanup

mikeoliphant:
This PR cleans up and simplifies the processing interface by doing the following:

- Removes input/output gain parameters.
- Removes model "params" dictionary parameter.
- Switches from multi-channel inputs/outputs to single mono input/output.
- Removes extra input/output buffers and input copy

The process call is now just simply:

`void DSP::process(NAM_SAMPLE* input, NAM_SAMPLE* output, const int num_frames)`

This mostly is just a benefit for simplicity of interface, although for simple models like LSTM, it *might* make for a small noticeable performance increase.

@sdatkinson - note that I just removed the model parameter dictionary from the process call. I didn't take out the internal code using it - I figured that is better done by you.

I've tested WaveNet and LSTM, but not ConvNet.

Resolves https://github.com/sdatkinson/NeuralAmpModelerCore/issues/77
Resolves https://github.com/sdatkinson/NeuralAmpModelerCore/issues/27
Gives us a start on https://github.com/sdatkinson/NeuralAmpModelerCore/issues/49

### DIFF
--- a/NAM/dsp.cpp
+++ b/NAM/dsp.cpp
@@ -35,6 +35,7 @@ DSP::DSP(const double loudness, const double expected_sample_rate)
 void DSP::process(NAM_SAMPLE* input, NAM_SAMPLE* output, const int num_frames)
 {
   this->_input_samples = input;
+  this->_output_samples = output;
   this->_num_input_samples = num_frames;
 
   this->_ensure_core_dsp_output_ready_();
@@ -61,15 +62,13 @@ void DSP::_get_params_(const std::unordered_map<std::string, double>& input_para
 
 void DSP::_ensure_core_dsp_output_ready_()
 {
-  if (this->_core_dsp_output.size() < _num_input_samples)
-    this->_core_dsp_output.resize(_num_input_samples);
 }
 
 void DSP::_process_core_()
 {
   // Default implementation is the null operation
   for (size_t i = 0; i < _num_input_samples; i++)
-    this->_core_dsp_output[i] = _input_samples[i];
+    this->_output_samples[i] = _input_samples[i];
 }
 
 void DSP::_apply_output_level_(NAM_SAMPLE* output, const int num_frames)
@@ -77,7 +76,7 @@ void DSP::_apply_output_level_(NAM_SAMPLE* output, const int num_frames)
   const double loudnessGain = pow(10.0, -(this->mLoudness - TARGET_DSP_LOUDNESS) / 20.0);
   const double finalGain = this->mNormalizeOutputLoudness ? loudnessGain : 1.0;
   for (int s = 0; s < num_frames; s++)
-    output[s] = (NAM_SAMPLE)(finalGain * this->_core_dsp_output[s]);
+    output[s] = (NAM_SAMPLE)(finalGain * _output_samples[s]);
 }
 
 // Buffer =====================================================================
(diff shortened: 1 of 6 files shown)


sdatkinson:
Awesome! Based on the description, this sounds super. I'll have a look when I get a chance.

NAM/dsp.h
```diff
@@ -107,7 +100,7 @@ class DSP
   virtual void _process_core_();
 
   // Copy this->_core_dsp_output to output and apply the output volume
```
sdatkinson, line 102: Nit: Since we're getting rid of `gain`, might be able to simplify this even more.

NAM/dsp.cpp
```diff
@@ -32,15 +32,15 @@ DSP::DSP(const double loudness, const double expected_sample_rate)
...
 }
 
-void DSP::process(NAM_SAMPLE** inputs, NAM_SAMPLE** outputs, const int num_channels, const int num_frames,
-                  const double input_gain, const double output_gain,
-                  const std::unordered_map<std::string, double>& params)
+void DSP::process(NAM_SAMPLE* input, NAM_SAMPLE* output, const int num_frames)
 {
-  this->_get_params_(params);
-  this->_apply_input_level_(inputs, num_channels, num_frames, input_gain);
+  this->_input_samples = input;
```
sdatkinson, line 37: Nit: I bet we can simplify this even more: I think that `process()` can be a pure virtual function.

NAM/dsp.cpp
```diff
@@ -60,38 +60,23 @@ void DSP::_get_params_(const std::unordered_map<std::string, double>& input_para
...
-  if (this->_core_dsp_output.size() < this->_input_post_gain.size())
-    this->_core_dsp_output.resize(this->_input_post_gain.size());
 }
 
 void DSP::_process_core_()
 {
   // Default implementation is the null operation
-  for (size_t i = 0; i < this->_input_post_gain.size(); i++)
-    this->_core_dsp_output[i] = this->_input_post_gain[i];
+  for (size_t i = 0; i < _num_input_samples; i++)
```
sdatkinson, line 70: Nit: yeah, I'm probably going to PR to put this as the default implementation of `process()` and get rid of `_process_core_()` entirely 🙂

NAM/dsp.cpp
```diff
@@ -60,38 +60,23 @@ void DSP::_get_params_(const std::unordered_map<std::string, double>& input_para
...
 
-void DSP::_apply_output_level_(NAM_SAMPLE** outputs, const int num_channels, const int num_frames, const double gain)
+void DSP::_apply_output_level_(NAM_SAMPLE* output, const int num_frames)
 {
   const double loudnessGain = pow(10.0, -(this->mLoudness - TARGET_DSP_LOUDNESS) / 20.0);
-  const double finalGain = this->mNormalizeOutputLoudness ? gain * loudnessGain : gain;
-  for (int c = 0; c < num_channels; c++)
-    for (int s = 0; s < num_frames; s++)
-      outputs[c][s] = (NAM_SAMPLE)(finalGain * this->_core_dsp_output[s]);
+  const double finalGain = this->mNormalizeOutputLoudness ? loudnessGain : 1.0;
```
sdatkinson, line 77: Hmm, I see.

Question: What do you think about having this class not be responsible for normalizing? It can provide information, but leave it up to the plugin to do anything about it?  I'm happy either way, but perhaps a tiny bit happier pushing it out of this library because of how it'd clean up all of the subclasses and make it a little simpler from a dev standpoint I think.

sdatkinson (approved):
LGTM! 👍🏻 

Thanks for this PR!

NAM/dsp.cpp
mikeoliphant: Yep, that was my preference from the beginning: https://github.com/sdatkinson/NeuralAmpModelerCore/pull/18#issuecomment-1493366512 😉 

I think it makes sense for the plugin to do it, since it will already be doing output gain adjustment - so no additional pass necessary.

mikeoliphant:
Overall, I agree that there is more simplification/restructuring that can/should be done. I avoided going too far at once in the interest of making some quick progress that could easily be merged.

Issue #79, opened 2023-10-06
Get rid of `DSP::_process_core_()` and `DSP::_apply_output_level_()`, make `process()` do what `_process_core_()` does


sdatkinson:
While reviewing #78 I realized that the `DSP` class's structure might be somewhat simplified by making `process()` do what `_process_core_()` does currently. In most cases, the existing `_process_core_()` implementations could become the implementations for `process()` directly with minimal modifications (e.g. add three lines to assign the pointers and buffer length before proceeding with the rest).

However, the current `_apply_output_level_()` is taking care of output level normalization; if we want this to remain part of this lbirary's responsibility, then the current class structure is a nice and DRY solution, but I can also see the justification of pushing this out of `DSP` and letting plugins using NAM take care of it themselves (getting the information they want via e.g. a new `DSP::GetLoudness()`.

I'm somewhat keen to do the latter (simplify this library and push out the gain adjustment to be the responsibility of the plugin) but want to raise it as an issue here to see if anyone has any strong opinions or thoughts if I'm missing something.

sdatkinson:
Related: #58

mikeoliphant:
👍 on pushing the loudness adjustment out to the plugin.

If we're in agreement, feel free to assign this to me and I'll put together a PR.

sdatkinson:
Sounds great! Thanks 🙂

PR #80, opened 2023-10-07
Remove _process_core() and output normalization

mikeoliphant:
Removed _process_core(). Pass input/output/num_frames where needed instead of caching in local variables. Removed output normalization. Added functions for getting model normalization factor in dB and linear gain.

I also removed the anti-pop code for ConvNet - it needs to be replaced with a pre-warm.

Resolves https://github.com/sdatkinson/NeuralAmpModelerCore/issues/79

### DIFF
--- a/NAM/dsp.h
+++ b/NAM/dsp.h
@@ -63,26 +63,19 @@ public:
   //   DSP subclass implementation.
   virtual void finalize_(const int num_frames);
   double GetExpectedSampleRate() const { return mExpectedSampleRate; };
-  bool HasLoudness() { return mLoudness != TARGET_DSP_LOUDNESS; };
-  void SetNormalize(const bool normalize) { this->mNormalizeOutputLoudness = normalize; };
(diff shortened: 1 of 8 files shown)


NAM/dsp.h
```diff
@@ -63,44 +63,26 @@ class DSP
   //   DSP subclass implementation.
   virtual void finalize_(const int num_frames);
   double GetExpectedSampleRate() const { return mExpectedSampleRate; };
-  bool HasLoudness() { return mLoudness != TARGET_DSP_LOUDNESS; };
-  void SetNormalize(const bool normalize) { this->mNormalizeOutputLoudness = normalize; };
+  bool HasNormalization() { return mLoudness != TARGET_DSP_LOUDNESS; };
```
sdatkinson, line 66: Nit: this'll work, but I'm going to mess w/ this a little bit. (I prefer "loudness" since I'm going to remove the "target" from this and let downstream users of this library decide for themselves how loud they want to aim for.

sdatkinson (approved):
Thanks for your patience. LGTM! 👍🏻

Issue #89, opened 2023-10-16
Move pre-warm to overridable method


mikeoliphant:
The pre-warm code might be leaner in a separate method that gets called after model initialization in get_dsp.

Then WaveNet (and I assume ConvNet) can override it to pre-warm over the receptive field and LSTM can do nothing.

Also, it would make it possible to avoid the default pre-warm behavior and do something else if you bypass get_dsp.

PR #90, opened 2023-10-17
Moved pre-warm to DSP and call it in get_dsp()

mikeoliphant:
This PR adds a prewarm() method to the DSP base class that gets called from get_dsp().

The default behavior is to process the model over "_prewarm_samples" samples. By default, this is 0 - so nothing happens for LSTM (which is what we want).

WaveNet and ConvNet both set "_prewarm_samples" to the size of the model's receptive field.

@sdatkinson I don't use ConvNet, so I'm not sure I got the receptive field correct - can you make sure to check it?

Resolves https://github.com/sdatkinson/NeuralAmpModelerCore/issues/89
Resolves https://github.com/sdatkinson/NeuralAmpModelerCore/issues/61

### DIFF
--- a/NAM/convnet.cpp
+++ b/NAM/convnet.cpp
@@ -109,7 +109,12 @@ convnet::ConvNet::ConvNet(const int channels, const std::vector<int>& dilations,
   this->_head = _Head(channels, it);
   if (it != params.end())
     throw std::runtime_error("Didn't touch all the params when initializing ConvNet");
+
+  _prewarm_samples = 1;
+  for (size_t i = 0; i < dilations.size(); i++)
+    _prewarm_samples += dilations[i];
 }
+
 
 void convnet::ConvNet::process(NAM_SAMPLE* input, NAM_SAMPLE* output, const int num_frames)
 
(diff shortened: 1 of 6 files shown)


NAM/convnet.cpp
```diff
@@ -109,8 +109,13 @@ convnet::ConvNet::ConvNet(const int channels, const std::vector<int>& dilations,
   this->_head = _Head(channels, it);
   if (it != params.end())
     throw std::runtime_error("Didn't touch all the params when initializing ConvNet");
+
+  _prewarm_samples = 1;
+  for (size_t i = 0; i < dilations.size(); i++)
+    _prewarm_samples += dilations[i];
```
sdatkinson, line 115: This should be something like `dilations[i] * (kernel_size[i] - 1)`

_However_, [the kernel size for the ConvNet is hard-coded to 2](https://github.com/sdatkinson/NeuralAmpModelerCore/blob/3152c0e397f52d177bd127e1089f57a39950eb86/NAM/convnet.cpp#L56) so technically this should be correct.

I'm happy to merge this as-is and grab it myself instead of pestering you for a revision.

sdatkinson (approved):
LGTM! 👍🏻

Issue #99, opened 2024-01-17
Get rid of `nam::DSP::finalize_()`


sdatkinson:
It's ugly being part of the public interface to `nam::DSP`? Perhaps it can be worked into the end of the `process()` routine?

PR #110, opened 2024-09-08
Remove `nam::DSP::finalize_()`

sdatkinson:
## Description

Breaking change. Following this, Code calling NAM should not do ([e.g.](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/83fa931886f2fd0ab15e38d4bd5c5a9b23716717/NeuralAmpModeler/NeuralAmpModeler.cpp#L308)):

```cpp
model->process(input, output, numFrames);
model->finalize_(numFrames);
```

and instead just do

```cpp
model->process(input, output, numFrames);
```

And subclasses shall take care of any preparation for the next call to `process` themselves.

## Subclasses that don't override nor extend `DSP::finalize_()`:

* `Base`
* `Gain`
* `History`
* `LSTM`

## Subclasses where some change was needed:

* `Buffer`: This advanced the input buffer. Made an equivalent protected method `_advance_input_buffer_()` that subclasses can still use.
* `WaveNet`: Move `_advance_buffers_()` inside of `process`.

Subclasses of `Buffer` also needed to be checked:

* `ConvNet`: doesn't override, so `_advance_input_buffer_()` is pulled inside of `process()`
* `Linear`: Same

## Testing
I checked a render before and after using [a typical WaveNet](https://tonehunt.org/arsafes/b486d24a-dcbc-4623-9cb0-ccd5b5570f87) and ensured that they're the same.

### DIFF
--- a/NAM/convnet.cpp
+++ b/NAM/convnet.cpp
@@ -132,6 +132,9 @@ void nam::convnet::ConvNet::process(NAM_SAMPLE* input, NAM_SAMPLE* output, const
   // Copy to required output array (TODO tighten this up)
   for (int s = 0; s < num_frames; s++)
     output[s] = this->_head_output(s);
+
+  // Prepare for next call:
+  nam::Buffer::_advance_input_buffer_(num_frames);
 }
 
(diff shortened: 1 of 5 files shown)


sdatkinson:
Resolves #99

PR #120, opened 2024-10-10
CI: Test loading and running models

### LINKED ISSUE #47: CI: Test loading and running various models
Thanks to #34 it's possible to build some tools automatically 🎉 

Next step would be to try running them with some models (LSTM, WaveNet, ConvNet).

Some super-light models should do--the goal I have with this Issue is to be able to assert correct operation--so we can be more confident as we iterate on the core code that we're not breaking models, etc.

sdatkinson:
WaveNet and LSTM as the main cases to cover; more can come later

Resolves #47


.github/workflows/build.yml
```diff
@@ -22,3 +22,9 @@ jobs:
       run: |
         cmake .. -DCMAKE_BUILD_TYPE=$BUILD_TYPE
         cmake --build . --config $BUILD_TYPE -j4
+    
+    - name: Run tests
+      working-directory: ${{github.workspace}}
+      run: |
+        ./build/tools/benchmodel ./example_models/wavenet.nam
```
sdatkinson, line 29: It's not the best, but we'll iterate

PR #134, opened 2024-11-05
update nlohmann json to latest to be able to compile NeuralAmpModelerCore with C++20 and also C++23

erikgrahn13:
In the current state of the main branch of NeuralAmpModelerCore you cannot compile it with C++20 on MacOS. With MSVC it seems to still compile with C++20 but using the XCode toolchain it breaks due to the old version of nlohmann json


sdatkinson:
Superseded by #152. Thanks!

Issue #140, opened 2025-03-03
[BUG] Fix the build


sdatkinson:
Checks failed during #139:

https://github.com/sdatkinson/NeuralAmpModelerCore/actions/runs/13623308113/job/38076330519?pr=139

Get the build working again.

sdatkinson:
Re-trying on commit [e181f61](https://github.com/sdatkinson/NeuralAmpModelerCore/commit/e181f61efb8d05d34add45b5eecb3893ff21177c), which worked when it was merged: https://github.com/sdatkinson/NeuralAmpModelerCore/actions/runs/13623430545.

Failed. Hmm.

PR #141, opened 2025-03-08
[BUGFIX] Eliminate real-time allocations in WaveNet

### LINKED ISSUE #49: Do not allocate memory during WaveNet processing

sdatkinson:
* Non-gating WaveNet should be real-time safe (benchmodel with example WaveNet runs clean).
* Some tests would be great though.

Sort-of breaking changes (causing me to bump to 0.3.0):
* `Reset()` calls `prewarm()`. (This is now ok because it calls `SetMaxBufferSize()` first, which can handle allocation of model arrays.)
* `mMaxBufferSize` defaults to `0` instead of `512` on construction. This is nice and conservative but might break code that was expecting teh mdoel to come "ready to go" out of `get_dsp()` (However, `Reset()` _should_ be called before using the model!)

Resolves #49


sdatkinson:
Ugh, I need to get #140 sorted. I'll get that figured out before I release 0.3.0.

PR #147, opened 2025-07-23
Delete Warnings compile on clang macOS

proboofficial:
```
#ifndef __APPLE__
#include <ciso646> // and, not, or
#endif
```

I add this for elimination warning clang macOS compile for every include ciso646


Dependencies/.DS_Store
sdatkinson, line 1: Please remove .DS_Store files

Dependencies/nlohmann/json.hpp
```diff
@@ -36,7 +36,9 @@ SOFTWARE.
 
 #include <algorithm> // all_of, find, for_each
 #include <cassert> // assert
+#ifndef __APPLE__
```
sdatkinson, line 39: What's defining `__APPLE__`? It's something outside of the scope of this repo?

If you can help explain to me what the issue is (ideally, point me to an issue on the repo that this comes from!) then that'll help me understand whether it's a good idea to deviate from nlohmann's code.

sdatkinson (changes requested):
See comments

Issue #165, opened 2025-11-29
Better testing framework


sdatkinson:
See #164.

Catch2 seems reasonable. Opening this Issue to invite discussion. cc @Khalian

Khalian:
#164 is a bit premature. I am redoing this work, and will most likely end up with a new PR. 

So Catch2 is reasonable and the reason I proposed it 3 fold

1. Setting up a testing framework with proper help functionality, test enumeration etc.
2. Natively supporting benchmarking idiomatically. 
3. Setup a way to check for memory leaks (e.g. using valgrind) against the tests on catch2. 

Catch2 happens to be the simplest way to do C++ tests and benchmarking idiomatically

PR #166, opened 2025-12-01
Migrate to using catch2 as test framework

Khalian:
Migrating to using catch2 as a framework. 

Solves issue https://github.com/sdatkinson/NeuralAmpModelerCore/issues/165. 

However at a later date, I would like to move benchmodel to also use catch2 and setup valgrind memory leak checks as well. Those would be different PRs.


Khalian:
@sdatkinson Pinging here. This might have fallen off of your radar.

PR #180, opened 2026-01-14
Added gating activation classes

jfsantos:
Adding abstractions for gating activations and blending activations. These classes are both very similar: they take two activations (and a blending coefficient alpha for blending activations). One activation is used for channel 0 of the input (the "input activation" in the code) and the other is used for channel 1 (the "gating activation" in the code). The difference between the two is that gating activation uses the activation of the second channel as a gate, while blending activation sums both channels scaled by alpha and (1-alpha).

### DIFF
--- a/NAM/gating_activations.h
+++ b/NAM/gating_activations.h
@@ -22,23 +22,19 @@ public:
   // Inherit the default apply methods which do nothing (linear/identity)
 };
 
-// Static instance for default activation
-static IdentityActivation default_activation;
-
 class GatingActivation
 {
 public:
   /**
    * Constructor for GatingActivation
-   * @param input_act Activation function for input channels (default: linear)
-   * @param gating_act Activation function for gating channels (default: sigmoid)
+   * @param input_act Activation function for input channels
+   * @param gating_act Activation function for gating channels
    * @param input_channels Number of input channels (default: 1)
    * @param gating_channels Number of gating channels (default: 1)
    */
-  GatingActivation(activations::Activation* input_act = nullptr, activations::Activation* gating_act = nullptr,
-                   int input_channels = 1)
-  : input_activation(input_act ? input_act : &default_activation)
-  , gating_activation(gating_act ? gating_act : activations::Activation::get_activation("Sigmoid"))
+  GatingActivation(activations::Activation* input_act, activations::Activation* gating_act, int input_channels = 1)
+  : input_activation(input_act)
+  , gating_activation(gating_act)
   , num_channels(input_channels)
   {
     assert(num_channels > 0);
@@ -102,10 +98,9 @@ public:
    * @param blend_act Activation function for blending channels
    * @param input_channels Number of input channels
    */
-  BlendingActivation(activations::Activation* input_act = nullptr, activations::Activation* blend_act = nullptr,
-                     int input_channels = 1)
-  : input_activation(input_act ? input_act : &default_activation)
-  , blending_activation(blend_act ? blend_act : &default_activation)
+  BlendingActivation(activations::Activation* input_act, activations::Activation* blend_act, int input_channels = 1)
+  : input_activation(input_act)
+  , blending_activation(blend_act)
   , num_channels(input_channels)
   {
     if (num_channels <= 0)
--- a/tools/test/test_gating_activations.cpp
+++ b/tools/test/test_gating_activations.cpp
@@ -24,7 +24,9 @@ public:
     Eigen::MatrixXf output(1, 3);
 
     // Create gating activation with default activations (1 input channel, 1 gating channel)
-    nam::gating_activations::GatingActivation gating_act(nullptr, nullptr, 1);
+    nam::activations::ActivationIdentity identity_act;
+    nam::activations::ActivationSigmoid sigmoid_act;
+    nam::gating_activations::GatingActivation gating_act(&identity_act, &sigmoid_act, 1);
 
     // Apply the activation
     gating_act.apply(input, output);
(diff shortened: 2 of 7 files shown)


NAM/gating_activations.h
```diff
@@ -0,0 +1,104 @@
...
+  void apply(Eigen::MatrixXf& input, Eigen::MatrixXf& output)
+  {
+    // Validate input dimensions
+    if (input.rows() < 2) {
+      throw std::invalid_argument("GatingActivation: input matrix must have at least 2 rows");
+    }
+    
+    // Ensure output has correct dimensions
+    if (output.rows() != 1 || output.cols() != input.cols()) {
+      output.resize(1, input.cols());
```
sdatkinson: We shouldn't allow resizing like this because it risks the code not being real-time safe

NAM/gating_activations.h
```diff
@@ -0,0 +1,104 @@
...
+    }
+    
+    // Apply activations to the two rows
+    Eigen::MatrixXf input_row = input.row(0);
```
sdatkinson: It looks like this would only work with two channels (rows). I'd like for it to support arbitrary numbers of input channels. Check out the WaveNet's current implementation in the Layer class

NAM/gating_activations.h
```diff
@@ -0,0 +1,104 @@
...
+    gating_activation->apply(gating_row);
+    
+    // Element-wise multiplication
+    output = input_row.array() * gating_row.array();
+  }
+
+private:
+  activations::Activation* input_activation;
+  activations::Activation* gating_activation;
+  activations::Activation default_activation; // Default activation for safety
```
sdatkinson: I think we can just yell at people instead of having a "safe" default :)

sdatkinson (changes requested):
Couple call-outs

NAM/gating_activations.h
```diff
@@ -0,0 +1,204 @@
...
+public:
+  /**
+   * Constructor for GatingActivation
+   * @param input_act Activation function for input channels (default: linear)
+   * @param gating_act Activation function for gating channels (default: sigmoid)
+   * @param input_channels Number of input channels (default: 1)
+   * @param gating_channels Number of gating channels (default: 1)
+   */
+  GatingActivation(activations::Activation* input_act = nullptr, activations::Activation* gating_act = nullptr,
+                   int input_channels = 1, int gating_channels = 1)
```
sdatkinson: Change to just `channels`, and let this be the number for both (and the number of output channels).

NAM/gating_activations.h
```diff
@@ -0,0 +1,204 @@
...
+  /**
+   * Constructor for BlendingActivation
+   * @param input_act Activation function for input channels
+   * @param blend_act Activation function for blending channels
+   * @param alpha_val Blending factor (0.0 to 1.0)
+   * @param input_channels Number of input channels
+   * @param blend_channels Number of blending channels
+   */
+  BlendingActivation(activations::Activation* input_act = nullptr, activations::Activation* blend_act = nullptr,
+                     float alpha_val = 0.5f, int input_channels = 1, int blend_channels = 1)
```
sdatkinson: Sorry, I mean that alpha is what's calculated by the blending activation, not a parameter (like how the multiplication factor in the gating activation comes out of the gating activation.

Then blend that base-activated array with its (pre-activation) inputs. This'll require a non-in-place activation...a copy for now will suffice I think.

NAM/gating_activations.h
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Get the number of output channels
+   */
+  int get_output_channels() const { return num_input_channels; }
+
+private:
+  activations::Activation* input_activation;
+  activations::Activation* blending_activation;
+  float alpha;
```
sdatkinson: And this won't be here anymore.

tools/test/test_fast_lut.cpp
```diff
@@ -5,35 +5,33 @@
 
 #include "NAM/activations.h"
```
sdatkinson: I assume this is all formatter, but lmk if not.

tools/test/test_gating_activations.cpp
```diff
@@ -0,0 +1,215 @@
...
+    // Apply the activation
+    gating_act.apply(input, output);
+
+    // Verify dimensions
+    assert(output.rows() == 1);
+    assert(output.cols() == 2);
+
+    std::cout << "GatingActivation custom activations test passed" << std::endl;
+  }
+
```
sdatkinson: Nit: Adding some tests for numerical accuracy would be nice.

NAM/gating_activations.h
jfsantos: fixed

NAM/gating_activations.h
```diff
@@ -0,0 +1,204 @@
...
+      input_activation->apply(input_block);
+
+      // Apply activation to blend channels
+      Eigen::MatrixXf blend_block = input.block(num_input_channels, i, num_blend_channels, 1);
+      blending_activation->apply(blend_block);
+
+      // Weighted blending
+      // For wavenet compatibility, we assume one-to-one mapping
+      assert(num_input_channels == num_blend_channels);
+      output.block(0, i, num_input_channels, 1) = alpha * input_block + (1.0f - alpha) * blend_block;
```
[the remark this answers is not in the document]
jfsantos: fixed

NAM/gating_activations.h
```diff
@@ -0,0 +1,204 @@
...
+      Eigen::MatrixXf input_block = input.block(0, i, num_input_channels, 1);
+      input_activation->apply(input_block);
+
+      // Apply activation to gating channels
+      Eigen::MatrixXf gating_block = input.block(num_input_channels, i, num_gating_channels, 1);
+      gating_activation->apply(gating_block);
+
+      // Element-wise multiplication and store result
+      // For wavenet compatibility, we assume one-to-one mapping
+      assert(num_input_channels == num_gating_channels);
```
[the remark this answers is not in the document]
jfsantos: fixed

NAM/gating_activations.h
```diff
@@ -0,0 +1,204 @@
...
+#include <stdexcept>
+#include "activations.h"
+
+namespace nam
+{
+namespace gating_activations
+{
+
+// Default linear activation (identity function)
+class LinearActivation : public nam::activations::Activation
```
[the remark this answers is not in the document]
jfsantos: Fixed

PR #181, opened 2026-01-14
[BREAKING] Conv1D manages its own ring buffer

sdatkinson:
This is a bit of a big one.

The main change is to get each `Conv1D` to hold a ring buffer inside it that's managed automatically. The main methods for working with it are now more idiomatic, I think:
* `SetMaxBufferSize()` (Should be called from `Reset() on modules that own convolution modules)
* `Process()` to process an input buffer
* `GetOutput()` to access the results of a `Process()` call.

This causes quite a bit of re-work in the WaveNet (and ConvNet) classes.
I also made sure that the WaveNet is real-time safe when using this new setup, including some new tests for real-time safety. I think that the checking here may be better than we had before, and it sniffed out a few spots where Eigen's block functionality might have been causing some blocks to be evaluated, causing (de)allocations during code that should have been real-time safe (i.e. WaveNet's `process()` method). Possibly related, I also saw that **The WaveNet model in this change is about 5-15% faster than current `main`** when this is built into REAPER and I try rendering a minimal project (just one track with an instance of NeuralAmpModelerPlugin, with all the "extras" disabled except the DC blocker). I checked with buffer sizes of 64 and 1024.

So, Conv1D now lives in its own source/header files (not `dsp.{h,cpp}`). There's also a pair of `ring_buffer` files for that part, but since it's a bit specialized for use by the Conv1D class it might make sense to put it inside the `conv1d` files.

There's also lots more tests for the ConvNet and WaveNet models. I still didn't look super-close at the ConvNet model (and it's not guaranteed RT-safe), but since it's not a very interesting model I'm not focusing much on it at the moment.

Lots of changes to the WaveNet implementation, but it's all below the public interfaces, amazingly (since `_Layer` and `_LayerArray` are meant to be private).
Some of the main ones to call out are that the outputs of various objects are now generally "owned" by said object and accessed via `GetOutput()`-like methods. WaveNet layers and layer arrays have two different inputs and outputs.
Also, since the convolution ring buffers are now encapsulated within the convolution objects that own them, there aren't any "start/end" arguments that need to be provided to the layer array and layer `Process()` calls--much cleaner and less error-prone, I hope.
This is also where I ended up being a bit shy on the use of "block" operations with Eigen, which seems to have been causing RT safety issues in intermediate stages of developing this. I wonder if they may have caused RT safety issues in earlier versions of this code that weren't caught. Regardless, the test are much better at catching them.

Finally, lots more tests--about 50 of them. Hopefully this improves the airtightness of things moving forward :) Most were AI-drafted, but I've read them all to ensure that they are sane, and I made some changes to get there :)

### DIFF
--- a/tools/benchmark_compare.sh
+++ b/tools/benchmark_compare.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# Script to compare performance of current branch against main
-# Usage: ./tools/benchmark_compare.sh [--model MODEL_PATH]
+# Script to compare performance of current branch against another branch (default: main)
+# Usage: ./tools/benchmark_compare.sh [--model MODEL_PATH] [--branch BRANCH_NAME]
 
 set -e  # Exit on error
 
@@ -9,6 +9,7 @@ MODEL_PATH="example_models/wavenet_a1_standard.nam"
 BUILD_DIR="build"
 BENCHMARK_EXEC="build/tools/benchmodel"
 NUM_RUNS=10
+COMPARE_BRANCH="main"  # Default branch to compare against
 # Report file will be set with timestamp in main()
 
 # Colors for output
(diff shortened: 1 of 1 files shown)


sdatkinson:
Waiting to merge this one until after #179  and #180

sdatkinson:
_Deep breath_

Issue #172, opened 2026-01-12
Grouped convolutions


sdatkinson:
Allow for grouped convolutions (cf. [PyTorch](https://docs.pytorch.org/docs/stable/generated/torch.nn.Conv1d.html)).

If not specified, default to `1`, which preserves the current functionality.

PR #183, opened 2026-01-15
[FEATURE] Grouped convolutions

sdatkinson:
Grouped convolutions in `Conv1D` using the `groups` argument.

Relevant to #172 , but need `Conv1x1` before resolving


sdatkinson:
`Conv1x1` needs this too. Next PR.

PR #189, opened 2026-01-19
Head 1x1 convolution

jfsantos:
Added a head1x1 convolution layer which processes the output of each layer contributions to the final model output. This updates Wavenet and LayerArray to take a Head1x1Params struct that is used to configure head1x1.

### DIFF
--- a/NAM/wavenet.h
+++ b/NAM/wavenet.h
@@ -33,7 +33,7 @@ public:
          const Head1x1Params& head1x1_params)
   : _conv(channels, gated ? 2 * bottleneck : bottleneck, kernel_size, true, dilation)
   , _input_mixin(condition_size, gated ? 2 * bottleneck : bottleneck, false)
-  , _1x1(bottleneck, channels, true)
+  , _1x1(bottleneck, channels, groups_1x1)
   , _activation(activations::Activation::get_activation(activation)) // needs to support activations with parameters
   , _gated(gated) 
   , _bottleneck(bottleneck)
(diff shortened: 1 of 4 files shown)


tools/test/test_wavenet/test_full.cpp
```diff
@@ -145,9 +147,10 @@ void test_wavenet_zero_input()
   const bool with_head = false;
   const int groups = 1;
   const int groups_1x1 = 1;
+  nam::wavenet::Head1x1Params head1x1_params(false, channels, 1);
```
sdatkinson: Nit: I prefer named variables over literals (`false`, `1`) for readability

sdatkinson (changes requested):
Pretty good. One crit on a `+=` that I think snuck in.

NAM/wavenet.h
```diff
@@ -13,17 +14,36 @@ namespace nam
...
 public:
   _Layer(const int condition_size, const int channels, const int bottleneck, const int kernel_size, const int dilation,
-         const std::string activation, const bool gated, const int groups_input, const int groups_1x1)
-  : _conv(channels, gated ? 2 * bottleneck : bottleneck, kernel_size, true, dilation, groups_input)
+         const std::string activation, const bool gated, const int groups_input, const int groups_1x1,
+         const Head1x1Params& head1x1_params)
+  : _conv(channels, gated ? 2 * bottleneck : bottleneck, kernel_size, true, dilation)
   , _input_mixin(condition_size, gated ? 2 * bottleneck : bottleneck, false)
-  , _1x1(bottleneck, channels, true, groups_1x1)
+  , _1x1(bottleneck, channels, true)
```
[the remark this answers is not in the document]
jfsantos: oops, not sure how that happened, I probably thought I was editing head_1x1 during early testing. fixed

tools/test/test_wavenet/test_full.cpp
jfsantos: updated most of this around the code. in cases where head1x1_active is false, I did not bother updating the 1 to head1x1_groups since it has no effect in the test so I kept a literal 1.

PR #192, opened 2026-01-20
Replaced manual gating/blending with Gating/BlendingActivation classes.

jfsantos:
This PR replaces the manual gating implementation in wavenet.cpp by either a GatingActivation or a BlendingActivation instance and makes those configurable.

I implemented backwards compatibility with older config files that expect gated to be boolean, and also implemented backwards-compatible constructors that can take gated as a boolean instead of the enum I created. We can remove those if not needed, but they are here for completeness. 

### DIFF
--- a/NAM/gating_activations.h
+++ b/NAM/gating_activations.h
@@ -82,8 +82,6 @@ public:
       output.block(0, i, num_channels, 1) = input_block.array() * gating_block.array();
     }
     
-    // Debug: Print first sample to verify gating is working
-    // std::cout << "GatingActivation debug: first output sample = " << output(0, 0) << std::endl;
   }
 
   /**
@@ -117,10 +115,8 @@ public:
   , blending_activation(blend_act)
   , num_channels(input_channels)
   {
-    if (num_channels <= 0)
-    {
-      throw std::invalid_argument("BlendingActivation: number of input channels must be positive");
-    }
+    assert(num_channels > 0);
+
     // Initialize input buffer with correct size
     // Note: current code copies column-by-column so we only need (num_channels, 1)
     input_buffer.resize(num_channels, 1);


NAM/wavenet.h
```diff
@@ -132,21 +183,46 @@ class LayerArrayParams
...
+  const GatingMode gating_mode;
   const bool head_bias;
   const int groups_input;
   const int groups_1x1;
   const Head1x1Params head1x1_params;
+  const std::string gating_activation;
+  const std::string blending_activation;
+  
+  // Backward compatibility constructor with boolean gated parameter
+  LayerArrayParams(const int input_size_, const int condition_size_, const int head_size_, const int channels_,
```
sdatkinson: I'm gonna refactor this so that all of the backward compatibility is in the factory, then all the constructors below that can be nice & fresh.

NAM/wavenet.cpp
```diff
@@ -58,29 +58,33 @@ void nam::wavenet::_Layer::Process(const Eigen::MatrixXf& input, const Eigen::Ma
...
-      // TODO Need to support other activation functions here instead of hardcoded sigmoid
-      activations::Activation::get_activation("Sigmoid")->apply(this->_z.block(bottleneck, i, bottleneck, 1));
-    }
-    this->_z.block(0, 0, bottleneck, num_frames).array() *=
-      this->_z.block(bottleneck, 0, bottleneck, num_frames).array();
-    _1x1.process_(_z.topRows(bottleneck), num_frames); // Might not be RT safe
+    // Use the GatingActivation class
+    // Extract the blocks first to avoid temporary reference issues
+    auto input_block = this->_z.leftCols(num_frames);
+    auto output_block = this->_z.topRows(bottleneck).leftCols(num_frames);
```
sdatkinson: This is a bit tricky to me...so the bottom rows are for the gate/blend op, and the top are both for the main activation's input _and_ the gating activation's output.

I'll make a note of this.

sdatkinson (commented):
Few things I might prefer done a little differently.

Mainly:
1. Leave backward compatibility only at the factory, remove it from the constructors so there's "only one way to do it."
2. Remove default "Sigmoid secondary activations, requiring them to be specified as well.

Honestly, these are a bit "nit"ty but I think that removing the old code paths now will maybe be worth it with a bit more clarity, focusing the backward compatibility into one spot.

I'll take care of 'em and get this merged here

sdatkinson:
Continuing this on #193

PR #194, opened 2026-01-20
Configurable activations

jfsantos:
Since some activations now require parameters, I updated the code for `get_activation` to take in either a string or a JSON object. In the former case it works as previously for backward compatibility, but when passed a JSON object it will now parse the object for activation parameters and initialize them properly. This is supported for all activations that support parameters in the codebase (leaky ReLU, PReLU, leaky hard tanh).

~~A potentially undesired update is how PReLU handles multiple channels when it is initialized with a single parameter. I updated the code to broadcast the parameter to all channels. Right now it's being done dynamically just for testing, but if the feature is not required and we just want it to fail when the input has more channels than the PReLU activation, we can do that too.~~


NAM/activations.cpp
```diff
@@ -31,6 +31,59 @@ nam::activations::Activation* nam::activations::Activation::get_activation(const
   return _activations[name];
 }
 
+nam::activations::Activation* nam::activations::Activation::get_activation(const nlohmann::json& activation_config)
```
sdatkinson: I can live with this for now, but I'd prefer something more typed than a "mystery JSON" being the input object. Something like the model factory/registry pattern feels like it might work here.

sdatkinson (commented):
Worried about deallocation. Will chat.

NAM/activations.h
```diff
@@ -226,20 +273,21 @@ class ActivationPReLU : public Activation
...
-    int actual_channels = matrix.rows();
-
-    // NOTE: check not done during runtime on release builds
-    // model loader should make sure dimensions match
-    assert(actual_channels == n_channels);
-
+    unsigned long actual_channels = static_cast<unsigned long>(matrix.rows());
+    
+    // Prepare the slopes for the current matrix size
+    std::vector<float> slopes_for_channels = negative_slopes;
```
sdatkinson: Worried that this might not be real-time safe.

tools/test/test_activations.cpp
```diff
@@ -195,4 +195,150 @@ class TestPReLU
...
+    assert(!config.negative_slope.has_value());
+    assert(!config.negative_slopes.has_value());
+
+    auto act = nam::activations::Activation::get_activation(config);
+    assert(act != nullptr);
+  }
+
+  static void test_all_simple_types()
+  {
+    // Test that all simple activation types work
```
sdatkinson: Nit: Looks like we'll have to remember to come back here if we add an activation. Would be nice to have something that automatically enumerates what's there.

sdatkinson (approved):

PR #212, opened 2026-01-28
[WIP] Optimizing grouped convolutions

jfsantos:
Optimizing grouped convolutions by pre-computing weight block indices and unrolling loops for common numbers of groups. Also added a performance benchamark for Conv1D and Conv1x1.

Other potential updates (still not implemented):
- [ ] Store weights as block-diagonal sparse matrix, do one matmul instead of G matmuls
- [ ] Templated convolutions with compile-time static shapes to leverage specific compile-time optimizations


jfsantos:
Optimizations to be included in a future PR

PR #219, opened 2026-02-06
Add profiling instrumentation for NAM building blocks

jfsantos:
Adds a profiling framework (NAM/profiling.h, NAM/profiling.cpp) with NAM_PROFILE_START()/NAM_PROFILE_ADD() macros and 14 timing categories. Supports both desktop (std::chrono) and ARM Cortex-M7 (DWT cycle counter) backends. Profiling is compile-time gated via -DNAM_PROFILING.

Instruments wavenet _Layer::Process() and _LayerArray::ProcessInner() with per-category timing, and adds profiling reset/print calls to the benchmodel tool.

### DIFF
--- a/NAM/profiling.h
+++ b/NAM/profiling.h
@@ -1,14 +1,17 @@
 #pragma once
 
-// Comprehensive profiling for NAM building blocks
+// Dynamic profiling registry for NAM building blocks
 // Enable with -DNAM_PROFILING
 //
 // Usage:
-//   1. Call nam::profiling::reset() before benchmark
-//   2. Run model processing
-//   3. Call nam::profiling::print_results() to display breakdown
-//
-// Categories cover all WaveNet operations including FiLM modulation.
+//   1. Register profiling types at file scope (static init):
+//        static int PROF_FOO = nam::profiling::register_type("Foo");
+//   2. Call nam::profiling::reset() before benchmark
+//   3. In hot path:
+//        NAM_PROFILE_START();
+//        // ... code ...
+//        NAM_PROFILE_ADD(PROF_FOO);
+//   4. Call nam::profiling::print_results() to display breakdown
 
 #ifdef NAM_PROFILING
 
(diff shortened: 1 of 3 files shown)


NAM/profiling.h
```diff
@@ -0,0 +1,153 @@
...
+    copies = 0;
+    setzero = 0;
+    ringbuf = 0;
+    condition = 0;
+    lstm = 0;
+    other = 0;
+  }
+
+  uint32_t total() const {
+    return conv1d + input_mixin + layer1x1 + head1x1 + rechannel + conv1x1 + activation + film + copies + setzero + ringbuf + condition + lstm + other;
```
sdatkinson: This feels fragile. Is there a way to "register" things or something?

NAM/wavenet.cpp
```diff
@@ -89,6 +90,8 @@ void nam::wavenet::_Layer::set_weights_(std::vector<float>::iterator& weights)
 
 void nam::wavenet::_Layer::Process(const Eigen::MatrixXf& input, const Eigen::MatrixXf& condition, const int num_frames)
 {
+  NAM_PROFILE_START();
```
sdatkinson: Yeah...I'm not super-hot on things like this sitting in the middle of the code. Is there a way to do this that doesn't add lines like this?

NAM/profiling.h
jfsantos: This whole implementation was targeted at the Wavenet and I didn't think of generalizing it, only covered anything that will be in A2... but you are right, it would be better if we just kept a dictionary of things we are measuring and reported all of them, just like a logging library.

tools/memory_usage.cpp
```diff
@@ -0,0 +1,611 @@
...
+    r.add_weights(out_ch); // _bias(out_ch)
+  r.add_buffers((size_t)out_ch * M); // _output(out_ch, M)
+  return r;
+}
+
+// ─── Conv1D ─────────────────────────────────────────────────────────────────
+
+// Conv1D stores kernel_size weight matrices (each out_ch x in_ch) or depthwise
+// vectors, plus a bias vector, a ring buffer, and an output buffer.
+static MemoryResult conv1d_memory(int in_ch, int out_ch, int kernel_size, bool bias, int dilation, int groups, int M)
```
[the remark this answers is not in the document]
jfsantos: M is buffer_size everywhere, thanks for catching it. I'll rename it all across the file.

PR #226, opened 2026-02-12
Add inline GEMM optimizations and general performance improvements

jfsantos:
Hand-optimized GEMM kernels for small matrices common in NAM models, gated by #ifdef NAM_USE_INLINE_GEMM with Eigen fallback. Includes:

- Specialized Conv1D kernels: fused 4x4 and 2x2 kernel_size=3, plus fully-unrolled paths for 2x2 through 8x8 channel configurations
- Conv1x1 inline specializations for all common size combinations
- FiLM inline path with 4-element loop unrolling
- GatingActivation/BlendingActivation inline paths
- Branchless hardswish, 4-element loop unrolling for all activations
- SiLU added to LUT enable/disable
- Ring buffer refactored to Eigen block operations
- memcpy replacements for pure copy operations in wavenet
- Optimized single-channel output path in WaveNet::process
- Buffer size benchmark tool (benchmodel_bufsize)


NAM/activations.cpp
```diff
@@ -214,8 +220,12 @@ void nam::activations::Activation::disable_lut(std::string function_name)
...
     _activations["Sigmoid"] = sigmoid_bak;
   }
+  else if (function_name == "SiLU")
+  {
+    _activations["SiLU"] = silu_bak;
+  }
   else
   {
-    throw std::runtime_error("Tried to disable LUT for a function other than Tanh or Sigmoid");
+    throw std::runtime_error("Tried to disable LUT for a function other than Tanh, Sigmoid, or SiLU");
```
sdatkinson: This is where I understood why L40 was a new feature and not a bug fix :)

NAM/activations.h
```diff
@@ -119,18 +117,12 @@ inline float swish(float x)
...
-  }
-  else
-  {
-    return x * (x + 3.0) / 6.0;
-  }
+  // Branchless implementation using clamp
+  // hardswish(x) = x * relu6(x + 3) / 6
+  //              = x * clamp(x + 3, 0, 6) / 6
+  const float t = x + 3.0f;
+  const float clamped = t < 0.0f ? 0.0f : (t > 6.0f ? 6.0f : t);
```
sdatkinson: Interesting if this really is better? I'd be surprised if a compiler wouldn't figure out that these are the same.

NAM/dsp.cpp
```diff
@@ -453,10 +454,250 @@ void nam::Conv1x1::process_(const Eigen::Ref<const Eigen::MatrixXf>& input, cons
...
   }
 
   if (this->_do_bias)
+  {
+#ifdef NAM_USE_INLINE_GEMM
+    const int out_ch = (int)get_out_channels();
+    float* __restrict__ output_ptr = _output.data();
+    const float* __restrict__ bias_ptr = this->_bias.data();
+
+    // Specialized paths for common small channel counts
```
sdatkinson: Nit: is it worth doing everything from 1 to 8?

NAM/ring_buffer.cpp
```diff
@@ -35,22 +35,10 @@ void RingBuffer::Write(const Eigen::MatrixXf& input, const int num_frames)
...
-  {
-    for (int row = 0; row < channels; ++row)
-    {
-      _storage(row, _write_pos + col) = input(row, col);
-    }
-  }
+  // Write the input data at the write position using Eigen block operations
+  // This is more efficient than element-by-element copy as it allows
+  // the compiler to vectorize the operation.
+  _storage.middleCols(_write_pos, num_frames).noalias() = input.leftCols(num_frames);
```
sdatkinson: I wonder why I did that...is there a chance that this isn't real-time safe? I need to check the tests...

tools/benchmodel_bufsize.cpp
```diff
@@ -0,0 +1,96 @@
+#include <iostream>
```
sdatkinson: Can you tell me (docstring?) the difference between this and benchmodel.cpp?

tools/benchmodel.cpp
```diff
@@ -17,74 +18,85 @@ double outputBuffer[AUDIO_BUFFER_SIZE];
 
 int main(int argc, char* argv[])
 {
-  if (argc > 1)
+  if (argc < 2)
   {
-    const char* modelPath = argv[1];
-
-    std::cout << "Loading model " << modelPath << "\n";
+    std::cerr << "Usage: benchmodel <model_path> [--no-fast-tanh]\n";
```
sdatkinson: Awesome!

tools/benchmodel.cpp
```diff
@@ -17,74 +18,85 @@ double outputBuffer[AUDIO_BUFFER_SIZE];
...
 
-      exit(1);
-    }
+  std::unique_ptr<nam::DSP> model;
+  model = nam::get_dsp(std::filesystem::path(modelPath));
 
-    size_t bufferSize = AUDIO_BUFFER_SIZE;
-    model->Reset(model->GetExpectedSampleRate(), bufferSize);
-    size_t numBuffers = (48000 / bufferSize) * 2;
+  if (model == nullptr)
```
sdatkinson: Why'd the change detector go nuts on this file? Is it a formatting thing?

jfsantos:
You were on the spot about the risk with non-contiguous operations. I added tests and made sure the inline ops all work on non-contiguous matrices now.

tools/benchmodel.cpp
jfsantos: I think a block changed place when I added the if for disabling fasttanh... but it doesn't make a lot of sense. Let me check.

sdatkinson (approved):
LGTM!

PR #227, opened 2026-02-13
Refactor factory into separate config parser and unified create_dsp() construction path

jfsantos:
Add typed config structs (LinearConfig, LSTMConfig, ConvNetConfig, WaveNetConfig) and parse_config_json() functions per architecture. Introduce ModelConfig variant, ModelMetadata, and create_dsp() for unified model construction independent of JSON parsing. Refactor get_dsp() to use the new unified path. Register Linear factory (was previously missing).

This is being done in order to support parsers for formats other than JSON more easily. Those parsers should NOT be a part of Core, at least for now. I will be creating a separate repo for the binary .namb format that leverages this refactor if it gets merged.

### DIFF
--- a/NAM/get_dsp.cpp
+++ b/NAM/get_dsp.cpp
@@ -173,6 +173,8 @@ std::unique_ptr<DSP> create_dsp(std::unique_ptr<ModelConfig> config, std::vector
 {
   auto out = config->create(std::move(weights), metadata.sample_rate);
   apply_metadata(*out, metadata);
+  // "pre-warm" the model to settle initial conditions
+  // Can this be removed now that it's part of Reset()?
   out->prewarm();
   return out;
 }
--- a/NAM/wavenet.cpp
+++ b/NAM/wavenet.cpp
@@ -592,11 +592,11 @@ nam::wavenet::WaveNetConfig nam::wavenet::parse_config_json(const nlohmann::json
   {
     nlohmann::json layer_config = config["layers"][i];
 
-    const int groups = layer_config.value("groups_input", 1);
-    const int groups_input_mixin = layer_config.value("groups_input_mixin", 1);
+    const int groups = layer_config.value("groups_input", 1); // defaults to 1
+    const int groups_input_mixin = layer_config.value("groups_input_mixin", 1); // defaults to 1
 
     const int channels = layer_config["channels"];
-    const int bottleneck = layer_config.value("bottleneck", channels);
+    const int bottleneck = layer_config.value("bottleneck", channels); // defaults to channels if not present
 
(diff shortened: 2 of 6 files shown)


NAM/get_dsp.cpp
```diff
@@ -146,62 +148,102 @@ std::unique_ptr<DSP> get_dsp(const nlohmann::json& config, dspData& returnedConf
...
+// =============================================================================
+// Unified construction path
+// =============================================================================
+
+ModelConfig parse_model_config_json(const std::string& architecture, const nlohmann::json& config, double sample_rate)
 {
-  bool have = false;
-  double value = 0.0;
-};
+  if (architecture == "Linear")
```
sdatkinson: [C] It appears that the registry doesn't work anymore with this?

If I have another repo with a new NAM architecture, it won't get picked up anymore.

NAM/model_config.h
```diff
@@ -0,0 +1,51 @@
...
+{
+  std::string version;
+  double sample_rate = -1.0;
+  std::optional<double> loudness;
+  std::optional<double> input_level;
+  std::optional<double> output_level;
+};
+
+/// \brief Variant of all architecture configs
+using ModelConfig = std::variant<linear::LinearConfig, lstm::LSTMConfig, convnet::ConvNetConfig, wavenet::WaveNetConfig>;
```
sdatkinson: [C] Can this be added to by code outside of this repo?

NAM/wavenet.cpp
```diff
@@ -568,41 +568,43 @@ void nam::wavenet::WaveNet::process(NAM_SAMPLE** input, NAM_SAMPLE** output, con
...
   }
-  std::vector<nam::wavenet::LayerArrayParams> layer_array_params;
+
   for (size_t i = 0; i < config["layers"].size(); i++)
   {
     nlohmann::json layer_config = config["layers"][i];
 
-    const int groups = layer_config.value("groups_input", 1); // defaults to 1
-    const int groups_input_mixin = layer_config.value("groups_input_mixin", 1); // defaults to 1
+    const int groups = layer_config.value("groups_input", 1);
```
sdatkinson: Claude seems to have thrashed the comments, which is making it hard for my eyes to see the changes. Can you revert those?

I _think_ this is correct but

sdatkinson (changes requested):
This is pretty slick overall!

The big problem, however, is that it seems to kill the registry pattern that makes this library extensible with new architectures?

Perhaps I owe you a test that compile-time registers a new "test" architecture. Let me do that so that the requirement is more clear.

Otherwise, very cool!

PR #236, opened 2026-02-27
Fix MSVC build ("__restrict__" keyword)

mikeoliphant:
The latest NAM Core doesn't build with Visual Studio because the "__restrict__" keyword is different ("__restrict").

This fixes things with an MSVC-specific #define.


sdatkinson (approved):
Thanks!

PR #243, opened 2026-03-12
[WIP] First draft of SlimmableWavenet

jfsantos:
This is a bit of a big one. Most of the logic is behind parsing the weights to slim the model according to the config. Might be interesting to add a test to compare a non-slimmable model with a slimmed model where both have the same parameters and make sure they match, but I haven't done that yet. Submitting the PR just so you can get a head start on checking the API and the logic. 


sdatkinson:
I haven't looked at the implementation yet but I'm going to pre-empt that we may be able to make use of numerical integration tests (see https://github.com/Atkinson-Advanced-Modeling/NamIntegrationTests/pull/2) to ensure that the results are as we need them to be.

I'll keep that in mind as I review--let's PR checks there next.

sdatkinson (approved):
LGTM! 👍🏻

sdatkinson:
Resolves #241

PR #264, opened 2026-05-03
Bump Eigen submodule to 5.0.1 (closes #67)

### LINKED ISSUE #67: Ensure Eigen members are aligned correctly for optimized CPU instructions
See http://eigen.tuxfamily.org/dox-3.2/group__TopicStructHavingEigenMembers.html and act accordingly.

rhaist:
## Summary

Bumps `Dependencies/eigen` to the released **Eigen 5.0.1** tag.

- Submodule moves from `87300c93c` (pre-3.5 dev snapshot, 2023-04-17) → `bc3b39870` (5.0.1, 2025-11-11).
- One-line diff. No NAM source changes.

## Why this closes #67

Issue #67 asks the project to follow [Eigen's struct-having-Eigen-members guidance](http://eigen.tuxfamily.org/dox-3.2/group__TopicStructHavingEigenMembers.html) so consumers don't hit alignment crashes under aggressive vectorization, and so the README "Sharp edges" workarounds (`EIGEN_MAX_ALIGN_BYTES=0`, `EIGEN_DONT_VECTORIZE`) become unnecessary.

The Eigen 5.0 release notes resolve this at the library level. From `CHANGELOG.md`:

> Eigen now uses the c++11 **alignas** keyword for static alignment. Users targeting C++17 only and recent compilers (e.g., GCC>=7, clang>=5, MSVC>=19.12) will thus be able to completely forget about all issues related to static alignment, including `EIGEN_MAKE_ALIGNED_OPERATOR_NEW`.

NAM already requires C++20 (`CMakeLists.txt`). With Eigen 5.0, no per-class macros, no `aligned_allocator` audit, and no `EIGEN_MAX_ALIGN_BYTES` overrides are needed for correct alignment of either fixed-size or dynamic-size Eigen members. The README "Sharp edges" section can be retired in a follow-up.

I also briefly explored an alternative PR adding `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` to every class with Eigen members. Decided against it: today every member is dynamic-size (`MatrixXf`/`VectorXf`), so the macro is a runtime no-op, and Eigen 5.0 obsoletes it anyway. The submodule bump fixes the underlying problem rather than working around it.

## Why the 5.0.1 patch release

Pinning the latest patch release. The [5.0.1 release notes](https://gitlab.com/libeigen/eigen/-/releases/5.0.1) (2025-11-11) include two fixes directly relevant to NAM:

- **Alignment issues** (`#2982`, `#2984`) — alignment is the entire subject of this PR / issue #67.
- **Missing C++20 `<version>` header** (`#2986`) — NAM is a C++20 build, so this could otherwise bite the compile on some toolchains.

The remaining 5.0.1 fixes (BLAS/LAPACK Windows build `#2980`, `geo_homogeneous` tests `#2977`, dirty-git-state `#2995`) don't touch anything NAM uses.

## What I tested locally

Platform: macOS (Darwin 25.x, AppleClang 21.0.0), C++20, Apple Silicon.

Reproduced the same matrix CI runs on Ubuntu (`build.yml`), for **both** the upstream-`main` Eigen baseline and Eigen 5.0.1, with both GEMM paths:

| Variant | Configure | Compile | `run_tests` |
|---|---|---|---|
| `Debug`, default GEMM | clean | clean, no new warnings | Success |
| `Debug`, `-DNAM_USE_INLINE_GEMM` | clean | clean, no new warnings | Success |

The only warnings emitted in the new build are the three pre-existing `-Wsign-compare` warnings in `Dependencies/AudioDSPTools/dsp/wav.cpp` (unrelated, present on `main`).

### Render bit-exactness check

I rendered `example_models/wavenet.nam` over `example_audio/input.wav` (96000 samples) and compared SHA-256 hashes. Unlike the earlier revision of this PR, the **baseline was re-rendered on this same machine/compiler** from the current upstream-`main` Eigen pin (`before-3.4-780-g87300c93c`), rather than quoting prior numbers — so this is a true apples-to-apples comparison against the version actually being replaced.

| Variant | SHA-256 | |
|---|---|---|
| upstream `main` Eigen, default GEMM | `1d920928a7588f0455f8df088df3758d4f14e8afb4a3f09d5f9afdd477c533d9` | baseline |
| upstream `main` Eigen, inline GEMM | `465e0cb0c8be01f21aa570dce41fb754c0016abcf17dbfeac247fc195a9e84d1` | baseline |
| Eigen 5.0.1, default GEMM | `be598b753d127d9c90e5cee8bb59235f16f3e5fd294c2cc2c6110968a6fc4470` | ⚠️ differs from baseline |
| Eigen 5.0.1, inline GEMM | `465e0cb0c8be01f21aa570dce41fb754c0016abcf17dbfeac247fc195a9e84d1` | ✅ identical to baseline |

**Inline GEMM is bit-exact** across the bump (`main` → 5.0.1) — expected, since that path bypasses Eigen's matrix product. **Default Eigen GEMM is not bit-exact** — Eigen 5.0 changes product evaluation/vectorization, producing different LSB rounding. `run_tests` still passes on both Eigen versions because the unit-test tolerances are wider than the rounding noise.

How large is the default-GEMM change, numerically? Comparing the two `render` outputs sample-by-sample (96000 samples, peak signal 0.071):

| Metric | `main` Eigen → 5.0.1, default GEMM |
|---|---|
| max abs diff | `9.5e-08` |
| mean abs diff | `5.7e-09` |
| `np.allclose(rtol=1e-5, atol=1e-6)` | **True** |

So the change is pure LSB float rounding — about **10× below** the `atol=1e-6` used downstream.

### Integration tests are unaffected (no goldens to refresh)

The integration workflow (`Atkinson-Advanced-Modeling/NamIntegrationTests`) does **not** pin bit-exact golden audio. `test/test_numerical_agreement.py` exports a fresh trainer (PyTorch) snapshot at test time and asserts the core `render` output agrees within tolerance:

```python
_RTOL = 1e-5
_ATOL = 1e-6
assert np.allclose(actual_flat, expected_flat, rtol=_RTOL, atol=_ATOL)
```

Since the Eigen-3.x → 5.0.1 default-GEMM difference (max `9.5e-8`) is ~10× under that `atol`, these tests pass unchanged — there are no stored Eigen-version goldens that need updating. (The earlier revision of this PR speculated they might; that was overcautious.)

## What I did NOT verify locally

I'm on macOS, so I could not exercise the Linux side of the existing CI matrix. Specifically not run locally:

- `build-ubuntu` (libstdc++) — both default and `NAM_USE_INLINE_GEMM`
- `build-ubuntu-libcxx` (libc++ via `-DNAM_USE_LIBCXX_LINUX=ON`) — both default and `NAM_USE_INLINE_GEMM`. This job exists specifically to cover the `_LIBCPP_VERSION` branch in `wavenet/slimmable.h` (atomic shared_ptr free-functions vs `std::atomic<std::shared_ptr<…>>`).
- Windows / MSVC.

These will run on the existing `.github/workflows/build.yml` matrix when CI fires, which is the right place to catch any platform-specific Eigen 5 incompatibility.

## Other Eigen 5.0 changes worth noting

From the changelog, the changes most relevant to NAM:
- `Eigen::aligned_allocator` no longer inherits from `std::allocator` (C++ standard change). NAM does not use `aligned_allocator` directly — no impact.
- C++14 minimum for Eigen 5.0 (NAM is C++20 — no impact).
- New `EIGEN_ALIGNOF(X)` macro available if needed.
- `alignas`-based static alignment as quoted above.

## Suggested follow-ups (separate PRs, not in this one)

1. Drop the README "Sharp edges" section now that the workaround it documents is no longer needed.
2. Consider whether `Dependencies/AudioDSPTools/Dependencies/eigen` (still pinned at a 3.4.90 dev snapshot from 2023-09-14, but unused by NAM core — only `wav.cpp` is compiled and it doesn't include Eigen) should be bumped in `AudioDSPTools` for consistency.

## Test plan

- [ ] CI: `build-ubuntu` (libstdc++, default GEMM) — green
- [ ] CI: `build-ubuntu` (libstdc++, inline GEMM) — green
- [ ] CI: `build-ubuntu-libcxx` (libc++, default GEMM) — green
- [ ] CI: `build-ubuntu-libcxx` (libc++, inline GEMM) — green
- [ ] CI: integration tests — green (tolerance-based vs trainer output; no goldens to refresh — verified locally, max diff ~10x under atol)
- [x] Local macOS, default GEMM: `run_tests` + `benchmodel` + `render` — all pass (baseline + 5.0.1)
- [x] Local macOS, `NAM_USE_INLINE_GEMM`: `run_tests` + `benchmodel` + `render` — all pass (baseline + 5.0.1)
- [x] Render hash comparison upstream-`main` Eigen vs 5.0.1 documented above (baseline re-rendered on this machine)


sdatkinson:
Thanks for the very thoughtful and comprehensive PR :) Upgrading the linear algebra backend is very risky to me, so I appreciate the info.

PR #271, opened 2026-05-24
[PERFORMANCE] Templated kernels for grouped Conv1x1/Conv1D

### LINKED ISSUE #215: [PERFORMANCE] Grouped convolutions appear dominated by overhead
I'm profiling the models [here](https://drive.google.com/drive/folders/1geIj2Mi6V8qmCbLZXZo7n0SDWcxJR37s?usp=drive_link) and I'm getting poor results when grouped convolutions are used.

For example, `1x1_groups` looks at scaling upt he number of groups in an 8-channel 1x1:

Image

This plot makes it look like the overhead is what's dominating the calculation since the compute time scales linearly with the number of groups.

This is bad because it means that grouped convolutions are basically useless--better to just use the full matrix with a ton of zeroes.

I suspect that compile-time optimizations may solve this, but I'm surprised that it's _this_ bad.

rhaist:
Generalizes the depthwise-only fast path from #217 to all grouped (and small dense) `Conv1x1` / `Conv1D` shapes using compile-time-specialized kernels. Targets the "compile-time optimizations" path hinted at by #215.

Closes #215.

## Approach

`templated_conv1x1_kernel<OutCh, InCh, Groups>` and `templated_conv1d_tap_kernel<OutCh, InCh, Groups>` carry all shape information as template parameters. With constexpr loop bounds, the compiler unrolls every loop, folds index arithmetic, and never visits off-block-diagonal zeros.

Dispatch is a function pointer (`_kernel` / `_tap_kernel`) set at construction by `pick_*_kernel(out, in, groups)`. Unknown shapes return `nullptr` and fall through to the existing inline-GEMM / Eigen path — no regression risk. Both default Eigen and `NAM_USE_INLINE_GEMM` builds benefit (no `#ifdef` gate around dispatch).

Depthwise (`groups == channels`) is intentionally not registered — already handled by the existing `_is_depthwise` fast path from #217.

Registered square shapes: `(4,4)`, `(6,6)`, `(8,8)`, `(12,12)`, `(16,16)` at `groups in {1, 2, 3, 4, 6, 8}`.

## Microbenchmark (Conv1x1, 64-frame buffer, best of 3 x 2M iters)

| Shape | Baseline STD | Templated STD | Speedup |
|-------|--------------|---------------|---------|
| 8x8 G=1 | 167 ns | **69 ns** | 2.4x |
| 8x8 G=2 | 170 ns | **36 ns** | **4.7x** |
| 8x8 G=4 | 169 ns | **49 ns** | 3.5x |
| 16x16 G=2 | ~426 ns | **146 ns** | 2.9x |
| 16x16 G=4 | ~426 ns | **66 ns** | **6.5x** |
| 16x16 G=8 | ~426 ns | **97 ns** | 4.4x |

## End-to-end (`benchmodel`, best of 5 runs, STD build, Apple M-series Release)

`v4/1x1_groups/*.nam` (varies `layer1x1.groups`, 8-channel WaveNet):
| G | Baseline | Templated | Delta |
|---|----------|-----------|------|
| 1 | 9.97 ms | 8.76 ms | -12% |
| 2 | 9.83 ms | 8.29 ms | **-16%** |
| 4 | 9.91 ms | 8.55 ms | -14% |
| 8 | 8.99 ms | 9.13 ms | ~same (depthwise unchanged) |

`v4/input_groups/*.nam` (varies Conv1D `groups_input`):
| G | Baseline | Templated | Delta |
|---|----------|-----------|------|
| 1 | 7.42 ms | 6.66 ms | -10% |
| 2 | 6.97 ms | 5.62 ms | **-19%** |
| 4 | 6.98 ms | 5.41 ms | **-23%** |
| 8 | 6.66 ms | 5.53 ms | -17% |

`v4/channels/*.nam` (varies channel width, `groups=1` — shows dense gains from bypassing Eigen overhead on small matrices):
| ch | Baseline | Templated | Delta |
|----|----------|-----------|------|
| 4 | 17.21 ms | 12.39 ms | **-28%** |
| 8 | 29.36 ms | 25.11 ms | -14% |
| 12 | 45.71 ms | 38.40 ms | **-16%** |
| 16 | 65.50 ms | 65.04 ms | -1% |

Wins are independent of Eigen version — verified against both this repo's pinned Eigen 3.4-dev (`87300c93`) and a separately-tracked Eigen 5.0.0 bump.

## Correctness

| Check | Result |
|-------|--------|
| `run_tests` x {STD, NAM_USE_INLINE_GEMM} | pass |
| Conv1x1 templated vs reference dense GEMM (22 shape x groups, random weights) | max diff 1.2e-7 |
| Conv1D templated vs reference dense GEMM (198 shape x groups x K x dilation, random weights) | max diff 2.4e-7 |
| **Render parity on 33 production models** (`baseline a1-{pico,nano,feather,lite,standard}`, `channels/{1..16}`, `bottleneck_sizes`, `1x1_groups`, `input_groups`, `head1x1_groups`) | **bit-identical WAV output** (max diff = 0.0) |
| `git-clang-format --diff HEAD` | clean |
| New compiler warnings | none |
| Realtime allocations (raw-pointer kernel, no Eigen temporaries) | none |

## Diff

- `NAM/dsp.h` +10, `NAM/dsp.cpp` +120/-2
- `NAM/conv1d.h` +11, `NAM/conv1d.cpp` +132
- `tools/CMakeLists.txt` +36
- new `tools/bench_conv1x1_groups.cpp` (microbench + correctness gate)
- new `tools/check_conv1d_grouped.cpp` (Conv1D correctness gate across 198 shapes)

## Notes for reviewers

- `pick_*_kernel` tables are intentionally narrow — only square shapes that appear in the v4 model sweep. Trivial to extend later. Anything unregistered keeps the existing behavior exactly.
- The bench/check tools are not wired into `run_tests` because they need `-O2`/`-O3` (`run_tests` is `-O0` for allocation tracking). They run cleanly as standalone CI steps if you want them gated.

## Test plan

- [x] `run_tests` passes on STD + `NAM_USE_INLINE_GEMM`
- [x] `tools/bench_conv1x1_groups` correctness gate passes (22 shapes)
- [x] `tools/check_conv1d_grouped` passes (198 shapes)
- [x] Render parity on 33 production models: bit-identical
- [x] `git-clang-format --diff HEAD` clean

### DIFF
--- a/NAM/conv1d.cpp
+++ b/NAM/conv1d.cpp
@@ -4,6 +4,104 @@
 
 namespace nam
 {
+namespace
+{
+// Templated per-tap accumulating kernel for Conv1D.
+// OutCh, InCh, Groups are compile-time constants so the compiler unrolls every loop
+// and folds all index arithmetic. Off-block-diagonal zeros are never visited.
+// Weight memory layout is col-major (out_channels rows x in_channels cols), matching
+// Eigen::MatrixXf default storage in nam::Conv1D::_weight[k].
+// Input layout is assumed contiguous (channels rows x num_frames cols, col-major), as
+// the existing inline-GEMM cascade also assumes.
+template <int OutCh, int InCh, int Groups>
+void templated_conv1d_tap_kernel(const float* __restrict__ weight, const float* __restrict__ in,
+                                 float* __restrict__ out, int num_frames)
+{
+  static_assert(OutCh % Groups == 0, "OutCh must be divisible by Groups");
+  static_assert(InCh % Groups == 0, "InCh must be divisible by Groups");
+  constexpr int OutPerG = OutCh / Groups;
+  constexpr int InPerG = InCh / Groups;
+  for (int f = 0; f < num_frames; f++)
+  {
+    const float* __restrict__ in_col = in + f * InCh;
+    float* __restrict__ out_col = out + f * OutCh;
+    for (int g = 0; g < Groups; g++)
+    {
+      const int o_base = g * OutPerG;
+      const int i_base = g * InPerG;
+      for (int o = 0; o < OutPerG; o++)
+      {
+        float sum = 0.0f;
+        for (int i = 0; i < InPerG; i++)
+        {
+          sum += weight[(i_base + i) * OutCh + (o_base + o)] * in_col[i_base + i];
+        }
+        out_col[o_base + o] += sum;
+      }
+    }
+  }
+}
+
+// Map (out_channels, in_channels, groups) -> templated tap-kernel function pointer.
+// Returns nullptr for unregistered shapes; caller falls back to existing inline /
+// Eigen GEMM cascade. Depthwise (groups == channels) is handled by Conv1D's existing
+// _is_depthwise path and is intentionally not registered here.
+nam::Conv1D::TapKernel pick_conv1d_tap_kernel(int out_channels, int in_channels, int groups)
+{
+  using K = nam::Conv1D::TapKernel;
+  if (out_channels == 4 && in_channels == 4)
+  {
+    if (groups == 1)
+      return static_cast<K>(&templated_conv1d_tap_kernel<4, 4, 1>);
+    if (groups == 2)
+      return static_cast<K>(&templated_conv1d_tap_kernel<4, 4, 2>);
+  }
+  if (out_channels == 6 && in_channels == 6)
+  {
+    if (groups == 1)
+      return static_cast<K>(&templated_conv1d_tap_kernel<6, 6, 1>);
+    if (groups == 2)
+      return static_cast<K>(&templated_conv1d_tap_kernel<6, 6, 2>);
+    if (groups == 3)
+      return static_cast<K>(&templated_conv1d_tap_kernel<6, 6, 3>);
+  }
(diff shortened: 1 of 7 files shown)


rhaist:
Closing this one.

After reading through the [A2 complete guide](https://www.tone3000.com/guides/nam-a2-the-complete-guide), I realized this optimization targets the wrong path. The guide notes that A2 *explored grouped convolutions as one approach but ultimately prioritized simpler solutions when results were comparable*, achieving its efficiency instead through "fully unrolled, register-resident kernels sized to A2's channel counts" — i.e. the existing `NAM_ENABLE_A2_FAST` path in `a2_fast.{h,cpp}`.

Since A2 deliberately moved away from grouped convolutions, and A2-fast already bypasses the generic `Conv1D`/`Conv1x1` classes entirely, these templated grouped kernels would only accelerate a path that the current direction doesn't use. With A1 being deprioritized in favor of A2, the effort is better spent extending the unrolled A2-fast kernels (e.g. broader channel-count coverage, embedded ARM targets) than optimizing grouped convs.

Closing as not aligned with the A2 direction. Thanks!

PR #277, opened 2026-06-08
A2Fast: portable conv + memory optimizations (up to 1.6x on channels=8)

jfsantos:
Optimizes the A2-shape fast-path WaveNet (`a2_fast.cpp`) with portable, intrinsic-free changes. No change to model output; the generic WaveNet path is untouched.

### Changes
- **T=4 frame-tiled conv (channels ≥ 4).** Replaces the per-tap Eigen GEMM with a tap-major kernel that keeps a `T×Channels` accumulator local and reuses each weight column across 4 frames. The compiler emits SIMD broadcast-FMA (`vmlaq_n_f32` on AArch64, AVX equivalents on x86) — the same weight amortization as a hand-written microkernel, without intrinsics.
- **Direct head-history writes.** Layers write/accumulate activations straight into `_head_history` at the ring slot (`IsFirst` selects `=` vs `+=`), eliminating the separate `_head_sum` buffer, its per-block zeroing pass, and the head ring-write copy.
- **Dead terminal-residual elimination.** The final layer's `layer1x1` projection + residual store feed nothing downstream, so they're elided via an `IsLast` template parameter.
- **Single history arena.** All 23 layers' ring buffers come from one contiguous allocation, spreading base addresses across cache sets to reduce L1 set-conflict misses on narrow caches (e.g. Cortex-A8).
- **Mirror-on-demand ring writes.** Instead of a full `mbs×Channels` mirror refresh every block, predict each tap's next read range and mirror only the columns that actually overflow past `pow2_size` — frequently zero.
- **CMake:** default to `Release` when no build type is set (an empty `CMAKE_BUILD_TYPE` compiles unoptimized, ~10x slower for Eigen-heavy code).

### Benchmark (desktop, Apple Silicon, fast-path p50 µs/block, vs `main`)
Both builds forced to `-DCMAKE_BUILD_TYPE=Release` to isolate the code changes. The unchanged generic WaveNet ran as a control and stayed flat in both builds, confirming the deltas are real.

| buffer | a2lite (ch=3) | a2full (ch=8) |
|--------|---------------|---------------|
| 32 | 1.00x | 1.58x |
| 48 | 1.04x | 1.43x |
| 64 | 1.04x | 1.39x |
| 128 | 1.07x | 1.29x |

Channels=8 sees 1.3–1.6x from the tiled conv. Channels=3 is roughly flat on desktop; the arena/mirror changes target narrow embedded caches (Cortex-A8/M7) and aren't captured by a desktop run.

Some of these optimizations were contributed by [landon-chaos](https://github.com/landon-chaos), as he found them out while optimizing Core to run on his pedals :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,13 @@ project(NAM VERSION 0.4.0)
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
+# Default to a Release build: without this, an empty CMAKE_BUILD_TYPE compiles
+# with no optimization flags at all, which is 10x+ slower for Eigen-heavy code.
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+	set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type (Release, Debug, RelWithDebInfo, MinSizeRel)" FORCE)
+	message(STATUS "No build type specified; defaulting to Release")
+endif()
+
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED OFF)
 set(CMAKE_CXX_EXTENSIONS OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,11 +5,9 @@ project(NAM VERSION 0.4.0)
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
-# Default to a Release build: without this, an empty CMAKE_BUILD_TYPE compiles
-# with no optimization flags at all, which is 10x+ slower for Eigen-heavy code.
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
-	set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type (Release, Debug, RelWithDebInfo, MinSizeRel)" FORCE)
-	message(STATUS "No build type specified; defaulting to Release")
+	set(CMAKE_BUILD_TYPE Debug CACHE STRING "Build type (Release, Debug, RelWithDebInfo, MinSizeRel)" FORCE)
+	message(STATUS "No build type specified; defaulting to Debug")
 endif()
 
 set(CMAKE_CXX_STANDARD 20)
--- a/NAM/wavenet/a2_fast.cpp
+++ b/NAM/wavenet/a2_fast.cpp
@@ -150,7 +150,8 @@ private:
 
   void _load_weights(std::vector<float>& weights);
   void _ring_write(Layer& L, int num_frames);
-  void _layer_forward(int layer_idx, const float* cond, int num_frames, bool is_first, int head_wp);
+  void _layer_forward(int layer_idx, const float* cond, int num_frames, bool is_first, bool is_last,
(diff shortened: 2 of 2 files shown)


NAM/wavenet/a2_fast.cpp
```diff
@@ -578,53 +596,131 @@ void A2FastModel<Channels>::_layer_forward_k(Layer& L, const float* cond, int nu
...
+    // inner loop body is:
+    //
+    //   a[f][o] += Wcol[o] * h[f]   (o vectorized, h[f] scalar broadcast)
+    //
+    // where Wcol = W[:,cp] is stride-1 in o and h[f] = history[col+f][cp] is a
+    // scalar. On AArch64 the compiler emits vmlaq_n_f32 (SIMD FMA with scalar
+    // broadcast) with weight vector Wcol reused across all T frames — the same
+    // weight amortisation as an explicit NEON microkernel, without intrinsics.
+    // Equivalent SIMD broadcast-FMA instructions are emitted on x86 (AVX) and
+    // other SIMD targets automatically.
```
sdatkinson, line 615: neat :)

CMakeLists.txt
```diff
@@ -5,6 +5,13 @@ project(NAM VERSION 0.4.0)
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
+# Default to a Release build: without this, an empty CMAKE_BUILD_TYPE compiles
```
sdatkinson: I think I prefer defaulting to Debug--slower performance is very noticeable; losing your assertions when running tests, less so :)

sdatkinson (changes requested):
CMakeLists.txt but otherwise LGTM

sdatkinson (approved):
LGTM!

mikeoliphant:
Just a heads-up that this new optimized path seems to be very architecture and compiler dependent.

In my testing, it is significantly faster on x64 with clang, but slower on x64 with MSVC. On Raspberry Pi 4, it is much slower using both clang (v14) and gcc (v12).

jfsantos:
@mikeoliphant thanks for testing and reporting. I did not test with MSVC and on ARM I only tested in a couple of architectures (M7 and A8). Weird that I expected it to be faster on the RPi 4 but that isn't the case... did you compile with -mnative or cross-compile?

@sdatkinson Did you want me to gate these optimizations behind a compile time flag?

mikeoliphant:
@jfsantos I compiled natively on RPi4 (arm64), which enables the native optimizations by default using gcc. I don't know if clang does, but forcing "-march=native" doesn't improve things.

mikeoliphant:
I've now got confirmation the change also decreases performance on MOD Dwarf (ARM A35).

falkTX:
Bringing some benchmark data here from a ARMv8.2 board...

GCC9 without a2-fast:
```
Fast tanh: enabled
Loading model a2full.nam
Running benchmark
Finished
254ms
254.964ms
```

GCC15 without a2-fast:
```
Fast tanh: enabled
Loading model a2full.nam
Running benchmark
Finished
247ms
247.737ms
```

GCC15: with a2-fast:
```
Fast tanh: enabled
Loading model a2full.nam
Running benchmark
Finished
135ms
135.872ms
```

The improvement with a2-fast is quite drastic, sadly only makes sense in some compilers.

mikeoliphant:
I opened an issue (https://github.com/sdatkinson/NeuralAmpModelerCore/issues/290) to track this.

PR #281, opened 2026-06-10
Add load-time configuration options to `get_dsp()`

### LINKED ISSUE #269: Provide expected sample rate, max buffer size, and slimmable size as optional parameters to `get_dsp()`
Allow for specifying ☝🏻 when calling `get_dsp()` so that the model can be loaded directly as-expected without needing to call methods to re-configure (i.e. `Reset()` and `SetSlimmableSize()`).

Context: https://github.com/sdatkinson/NeuralAmpModelerCore/issues/266#issuecomment-4501789147 (and above)

sdatkinson:
Closes #269.

Adds optional get_dsp() parameters for caller-provided load-time configuration:

* expected sample rate
* max buffer size
* slimmable size

This lets callers load a model directly into the intended runtime state, avoiding extra post-load Reset() and SetSlimmableSize() calls. The default behavior is preserved when no options are provided.

Also updates slimmable model handling so initial slimmable selection can happen before reset/prewarm, and corrects the Reset() documentation to match current behavior.

### DIFF
--- a/NAM/container.cpp
+++ b/NAM/container.cpp
@@ -100,6 +100,13 @@ void ContainerModel::SetSlimmableSize(const double val)
   {
     return;
   }
+
+  if (!mHaveExternalSampleRate && GetMaxBufferSize() == 0)
+  {
+    _active_index.store(active_index, std::memory_order_release);
+    return;
+  }
+
   // Setting _active_index puts the model in the RT path, so reset before doing that.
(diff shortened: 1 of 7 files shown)


sdatkinson:
Current version of this I'm afraid doesn't resolve the actual issue. Continuing to massage.

sdatkinson:
If #285 turns out nicely, then I may change this and implement it by putting the defaults in a context.

PR #284, opened 2026-06-10
Make PrewarmSamples() public

### LINKED ISSUE #283: Have a public interface to get the model's receptive field size
Having external access to a model's receptive field is very useful in some cases.

The is currently a PrewarmSamples() method, but it is protected.

I would submit a PR that simply makes it public, but unfortunately things are a bit more complicated because container models return 0 (relying on prewarm() being run directly on the nested models).

If PrewarmSamples() for a container returned PrewarmSamples() for the active nested model, I *think* that would work.

mikeoliphant:
Also renamed it to GetPrewarmSamples() for clarity.

Container models now return active model GetPrewarmSamples()

Resolves https://github.com/sdatkinson/NeuralAmpModelerCore/issues/283

### DIFF
--- a/NAM/container.cpp
+++ b/NAM/container.cpp
@@ -108,6 +108,13 @@ void ContainerModel::SetSlimmableSize(const double val)
   _active_index.store(active_index, std::memory_order_release);
 }
 
+int ContainerModel::GetPrewarmSamples()
+{
+  const size_t active_index = _active_index.load(std::memory_order_acquire);
+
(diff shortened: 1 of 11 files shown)


mikeoliphant:
Looks like I missed a few files (ie: ConvNet) because I don't use them. Also some tests need updating.

@sdatkinson Can you patch that up if/when you merge?

mikeoliphant:
Nevermind - I fixed convnet and that was also the test issue.

sdatkinson (approved):
Thanks!

PR #285, opened 2026-06-10
[FEATURE] Options to disable prewarming during `.Reset()`

### LINKED ISSUE #270: Provide a method to temporarily suppress model prewarm behavior
It could be useful to have a way to temporarily suppress model prewarming if you want to take several actions (change slimmable size, set maximum buffer size, etc.) that would each individually prewarm the model.

Ideally, this would be possible to do before calling get_dsp(), but the model doesn't yet exist. Making it static would allow this, but might be undesirable.

sdatkinson:
## Description
Resolves #270

* `DSP::SetPrewarmOnReset()` to toggle whether reset triggers prewarming. Default to true to avoid mistakes; folks can use `false` to optimize their code (with care!).
* `ScopedPrewarmOnResetDefault` allows setting a context default before constructing DSP instances in case they call Reset internally during their constructor.
* Thread-local `gPrewarmOnResetDefault` is used to determine whether DSP instances prewarm on reset.

## Checklist
- [x] Verify that this can successfully suppress prewarming in NeuralAmpModelerPlugin's load path

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,19 @@
 cmake_minimum_required(VERSION 3.10)
 
-# Make sure this matches ./NAM/version.h!
-project(NAM VERSION 0.4.0)
+file(READ "${CMAKE_CURRENT_LIST_DIR}/NAM/version.h" _nam_version_header)
+foreach(_component MAJOR MINOR PATCH)
+	string(REGEX MATCH
+		"#define[ \t]+NEURAL_AMP_MODELER_DSP_VERSION_${_component}[ \t]+([0-9]+)"
+		_nam_version_match "${_nam_version_header}"
+	)
+	if(NOT _nam_version_match)
+		message(FATAL_ERROR "Could not parse NEURAL_AMP_MODELER_DSP_VERSION_${_component} from NAM/version.h")
+	endif()
+	set(_nam_version_${_component} "${CMAKE_MATCH_1}")
+endforeach()
+
+set(_nam_project_version "${_nam_version_MAJOR}.${_nam_version_MINOR}.${_nam_version_PATCH}")
+project(NAM VERSION "${_nam_project_version}")
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
--- a/NAM/container.cpp
+++ b/NAM/container.cpp
@@ -61,6 +61,13 @@ void ContainerModel::prewarm()
   _submodels[active_index].model->prewarm();
 }
 
+void ContainerModel::SetPrewarmOnReset(const bool prewarmOnReset)
+{
+  DSP::SetPrewarmOnReset(prewarmOnReset);
+  for (auto& submodel : _submodels)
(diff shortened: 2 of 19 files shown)


sdatkinson:
I've got this wired up in NeuralAmpModelerPlugin and have verified that I'm not visiting the prewarm routine more than once when changing models, regardless of whether the slim knob is set to 0 or 1.

So I'm feeling pretty good about this as a solution.

Code coming later in NeuralAmpModelerPlugin to demonstrate, but it's basically a no-prewarm context in `_StageModel` while the model is being initialized and packed into the resampling container; on exit, is `SetPrewarmOnReset(true)` and then finally `Reset()`. Since that's in the plugin code, I feel reasonably confident that other ways to build around `get_dsp` might be well-enough anticipated...so hopefully this does it!

sdatkinson:
Here's the method:

```cpp
std::string NeuralAmpModeler::_StageModel(const WDL_String& modelPath)
{
  WDL_String previousNAMPath = mNAMPath;
  try
  {
    auto dspPath = std::filesystem::u8path(modelPath.Get());
    std::unique_ptr<ResamplingNAM> temp = nullptr;
    {
      nam::ScopedPrewarmOnResetDefault scoped_prewarm_default(false);
      std::unique_ptr<nam::DSP> model = nam::get_dsp(dspPath);

      // Check that the model has 1 input and 1 output channel
      if (model->NumInputChannels() != 1)
      {
        throw std::runtime_error("Model must have 1 input channel, but has "
                                 + std::to_string(model->NumInputChannels()));
      }
      if (model->NumOutputChannels() != 1)
      {
        throw std::runtime_error("Model must have 1 output channel, but has "
                                 + std::to_string(model->NumOutputChannels()));
      }

      temp = std::make_unique<ResamplingNAM>(std::move(model), GetSampleRate());
      if (nam::SlimmableModel* slimmable = temp->GetSlimmableModel())
      {
        slimmable->SetSlimmableSize(GetParam(kSlim)->Value());
      }
    }
    temp->SetPrewarmOnReset(true);
    temp->Reset(GetSampleRate(), GetBlockSize());
    mStagedModel = std::move(temp);
    mNAMPath = modelPath;
    SendControlMsgFromDelegate(kCtrlTagModelFileBrowser, kMsgTagLoadedModel, mNAMPath.GetLength(), mNAMPath.Get());
  }
  catch (std::runtime_error& e)
  {
    SendControlMsgFromDelegate(kCtrlTagModelFileBrowser, kMsgTagLoadFailed);

    if (mStagedModel != nullptr)
    {
      mStagedModel = nullptr;
    }
    mNAMPath = previousNAMPath;
    std::cerr << "Failed to read DSP module" << std::endl;
    std::cerr << e.what() << std::endl;
    return e.what();
  }
  return "";
}
```

sdatkinson:
cc @mikeoliphant if you want to have a look before I merge this.

mikeoliphant:
So a thread-local flag? Seems like that should be enough to prevent the possibility of inter-plugin crosstalk.

PR #286, opened 2026-06-15
Support slimmable (SlimmableContainer) files in bench_a2_fast

jfsantos:
## What

`bench_a2_fast` previously only accepted plain `WaveNet` `.nam` files and bailed on slimmable files. A2 models are shipped as `SlimmableContainer` models that bundle multiple standalone submodels (e.g. A2-Full at 8 channels and A2-Lite at 3 channels), so they couldn't be benchmarked with this tool.

This adds a `--slim V` (0.0–1.0) option that resolves a `SlimmableContainer` to a single submodel using the same rule as `ContainerModel::SetSlimmableSize` (first submodel whose `max_value` exceeds `V`, else the last), then runs the existing fast-vs-generic comparison on that submodel's config + weights.

## Notes

- Plain `WaveNet` files are unaffected — passing `--slim` there just prints a note and benches normally.
- Omitting `--slim` on a container defaults to full size (last submodel).
- No library changes: each container submodel is already a standalone a2-shaped `WaveNet` spec, so it flows through the existing fast/generic paths directly.

## Testing

Ran against a real A2 `SlimmableContainer` file:

- `--slim 1.0` → submodel 1, A2 standard (8 ch), ~1.6× fast-vs-generic speedup
- `--slim 0.2` → submodel 0, A2 nano (3 ch), ~6.5× fast-vs-generic speedup

### DIFF
--- a/tools/bench_a2_fast.cpp
+++ b/tools/bench_a2_fast.cpp
@@ -6,7 +6,12 @@
 // Runs both over the same audio, reports median wall time + real-time factor.
 //
 // Usage:
-//   bench_a2_fast [--buffer N] [--seconds S] [--iters I] <model.nam> [<model.nam> ...]
+//   bench_a2_fast [--buffer N] [--seconds S] [--iters I] [--slim V] <model.nam> [<model.nam> ...]
+//
+// A model may be a plain WaveNet .nam or a SlimmableContainer .nam (e.g. an A2 file
+// bundling A2-Full + A2-Lite submodels). For a container, --slim V (0.0-1.0) selects the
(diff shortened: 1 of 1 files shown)


sdatkinson:
We can do this as a dynamic cast to the interface class, right? Looks like this is checking for the specific `SlimmableContainer` type of model.

Lmk if it can be done the way I'm thinking; if so, I'd prefer the tweak. Thanks!

sdatkinson:
Coming back to this. I think I understand why it's done this way--by the time we're casting `SlimmableModel`, it's too late to choose the DSP implementation.

Either way, the implementation seems like a two-way door so I'll merge it. Thanks!

PR #287, opened 2026-06-15
Fix GCC 9.4 build: gate std::atomic<shared_ptr> on C++20 feature macro

jfsantos:
## Problem

This project fails to build under GCC 9.4.0 because `NAM/wavenet/slimmable.h` uses `std::atomic<std::shared_ptr<T>>`, a C++20 library feature that **libstdc++** only supports from GCC 12.

The header already had a fallback to the deprecated `std::atomic_*` free-function overloads for `shared_ptr`, but it was keyed on `#ifdef _LIBCPP_VERSION`. GCC 9.4 uses **libstdc++** (not libc++), so it took the `#else` branch and required the unsupported `std::atomic<std::shared_ptr<>>`.

## Fix

Key the selection on the standard C++20 feature-test macro `__cpp_lib_atomic_shared_ptr` instead of on a specific standard library:

```cpp
#if defined(__cpp_lib_atomic_shared_ptr) && __cpp_lib_atomic_shared_ptr >= 201711L
  #define NAM_HAS_ATOMIC_SHARED_PTR 1
#else
  #define NAM_HAS_ATOMIC_SHARED_PTR 0
#endif
```

- **GCC 12+ (macro defined)** → unchanged, uses the member `std::atomic<std::shared_ptr<>>` path.
- **libc++ (macro never defined — feature still unimplemented)** → unchanged, uses the free-function fallback exactly as before.
- **GCC 9.4 + libstdc++ (macro undefined)** → now correctly takes the free-function fallback instead of failing to compile.

No behavior change for any compiler that previously built.

## Files changed

- `NAM/wavenet/slimmable.h` — feature-macro guard, conditional include, member declaration.
- `NAM/wavenet/slimmable.cpp` — the four staging accessor definitions.

### DIFF
--- a/NAM/wavenet/slimmable.cpp
+++ b/NAM/wavenet/slimmable.cpp
@@ -291,27 +291,7 @@ bool is_full_size(const std::vector<wavenet::LayerArrayParams>& params, const st
 
 } // anonymous namespace
 
-#ifdef _LIBCPP_VERSION
-void SlimmableWavenet::_pending_clear_release()
-{
-  std::atomic_store_explicit(&_pending_staged, std::shared_ptr<StagedSlimModel>{}, std::memory_order_release);
-}
-
-std::shared_ptr<SlimmableWavenet::StagedSlimModel> SlimmableWavenet::_pending_load_acquire() const
-{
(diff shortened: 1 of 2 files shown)


sdatkinson (approved):
LGTM 👍🏻

PR #299, opened 2026-06-25
Fix LSTM real-time safety (#218)

### LINKED ISSUE #218: Fix LSTM real-time safety issues
LSTM is not real-time safe.

Add tests similar to WaveNet and fix

rhaist:
## Summary

Fixes #218 — `LSTM` is not real-time safe.

`LSTMCell::get_hidden_state()` returned an `Eigen::VectorXf` by value, which heap-allocates on every call. Since the LSTM runs sample-by-sample on the audio thread, this allocated once per inter-layer hop and once for the head on **every single sample** (~48k × num_layers allocations/sec), making `process()` not real-time safe.

## Changes

- `get_hidden_state()` now returns a non-owning `Eigen::Ref<const VectorXf>` view into the internal state instead of a by-value copy.
- `LSTMCell::process_()` accepts the same `Eigen::Ref` so the view binds without copying.
- The gate matmul is split into a `noalias()` product plus a separate bias add, so the matrix-vector product evaluates directly into the pre-allocated `_ifgo` buffer with no temporary.

## Tests

Adds `tools/test/test_lstm_realtime_safe.cpp`, modeled on the existing WaveNet/FiLM real-time-safety tests, using the allocation-tracking harness to assert `LSTM::process()` performs **zero allocations and zero frees** across single-layer, multi-layer, multi-channel, large-hidden, and consecutive-call cases.

Verified the new tests catch the regression: **128 allocs / 128 frees before the fix, 0 / 0 after.** Full `run_tests` suite passes; `lstm.nam` runs clean end-to-end.

### DIFF
--- a/NAM/lstm.cpp
+++ b/NAM/lstm.cpp
@@ -28,14 +28,16 @@ nam::lstm::LSTMCell::LSTMCell(const int input_size, const int hidden_size, std::
     this->_c[i] = *(weights++);
 }
 
-void nam::lstm::LSTMCell::process_(const Eigen::VectorXf& x)
+void nam::lstm::LSTMCell::process_(const Eigen::Ref<const Eigen::VectorXf>& x)
 {
   const long hidden_size = this->_get_hidden_size();
   const long input_size = this->_get_input_size();
   // Assign inputs
   this->_xh(Eigen::seq(0, input_size - 1)) = x;
-  // The matmul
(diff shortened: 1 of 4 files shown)


sdatkinson (approved):
Thanks!

sdatkinson:
Thank you for the bunch of helpful PRs! Much appreciated 😄

PR #300, opened 2026-06-25
Fix A2 fast-path prewarm count to match the generic WaveNet

rhaist:
## Summary

`A2FastModel` (the A2 fast-path WaveNet, a drop-in for the generic `WaveNet`)
computed its prewarm sample count as the layer-stack **lookback distance** —
`Σ (kernel_size-1)·dilation + (head_kernel-1)` — instead of the **receptive
field**, which is one greater. The generic path seeds `mPrewarmSamples` at `1`
(the sample being produced) before adding the same lookback terms
(`model.cpp`), so the fast path warmed up by **one fewer sample** than the
model it replaces.

## Impact

`prewarm()` runs `process()` in whole `maxBufferSize` blocks until the count is
reached, so the off-by-one only changes the block count when the total (`6346`
for the A2 shape) is an exact multiple of the buffer size. Power-of-2 buffers
(64/256/4096) mask it — which is exactly why the existing equivalence test
(`test_matches_generic`, blocks 64/256) never caught it. On a buffer size that
divides `6346` (e.g. 334), the two paths warm up by a different number of
blocks and their first post-`Reset` output diverges.

Low severity (no audible artifact at typical buffer sizes, no RT-safety
impact), but a real contract violation for a path whose whole purpose is to be
numerically identical to the generic WaveNet.

## Fix

- Seed `prewarm` at `1` to match the generic receptive-field formula, with a
  comment explaining the anchor so the count isn't re-derived incorrectly again.
- Add `test_prewarm_matches_generic_{nano,standard}`: builds both the fast and
  generic model from the same config and asserts equal `GetPrewarmSamples()`.

## Verification

Confirmed the test catches the regression — reverting the seed to `0` aborts
the new assertion; with the fix the full `run_tests` suite passes (A2 fast path
is built by default via `NAM_ENABLE_A2_FAST`).

### DIFF
--- a/NAM/wavenet/a2_fast.cpp
+++ b/NAM/wavenet/a2_fast.cpp
@@ -172,7 +172,11 @@ A2FastModel<Channels>::A2FastModel(std::vector<float> weights, double expected_s
 
   _load_weights(weights);
 
-  int prewarm = 0;
+  // Receptive field = 1 (the sample being produced) + sum of per-layer lookbacks +
+  // (head kernel - 1). The leading 1 matches the generic WaveNet's prewarm count
+  // (model.cpp: mPrewarmSamples starts at 1 when there's no condition DSP), so the
+  // fast path warms up by exactly the same number of samples as the model it replaces.
+  int prewarm = 1;
   for (int i = 0; i < kNumLayers; i++)
     prewarm += _layers[i].max_lookback;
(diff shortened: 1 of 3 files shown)


sdatkinson (approved):
Thanks again!

PR #302, opened 2026-06-27
Bumped NEURAL_AMP_MODELER_DSP_VERSION_PATCH

benediktadams:
I'm assuming that this should be in sync with the repo version tag at commit 1f42f8... please disregard if I'm misunderstanding the git version tags.


sdatkinson:
Ach, you're right .

Mind moving it to 0.5.5? I prefer for the tip of `main` to look like the "next" release instead of "also the most recent release" and 0.5.4 has already been released.

(I should also make some automation for this repo to handle this for me so I stop making this mistake :) )

PR #311, opened 2026-08-02
Validate .nam model config fields instead of unchecked operator[]

constk:
## Problem

`.nam` files are untrusted input — hosts download them and open them on a user's behalf. The config parsers read required fields with `config["key"]`, which resolves to nlohmann's **const** `operator[]`. That does `JSON_ASSERT(key exists)` and then dereferences the iterator:

```cpp
// Dependencies/nlohmann/json.hpp:22182-22193
JSON_ASSERT(it != m_data.m_value.object->end());
return it->second;
```

`JSON_ASSERT` is plain `assert` (`json.hpp:2571`). Under `NDEBUG` — i.e. any release plugin build — the assert vanishes and a model file missing a required key **dereferences a past-the-end iterator** instead of throwing. That's undefined behaviour reachable from a downloaded file.

Alongside it, several attacker-controlled values reached allocations and arithmetic unchecked:

- Dimensions were read as raw `int` and passed to `Eigen::resize`. A negative value becomes a huge `size_t`.
- `groups` was never validated, and reaches `if (in_channels % groups != 0)` in `dsp.cpp:314` / `conv1d.cpp:60,65`. **`"groups_input": 0` is a one-field divide-by-zero.**
- Array lengths were unbounded, so a small, highly compressible file could ask for millions of `Layer` objects.
- Non-integral numbers were silently truncated (`"channels": 4.9` → `4`).

## Approach

New `nam::util` helpers in `NAM/json_util.h` look a field up, name both the offending key and its enclosing context in the message, and throw `std::runtime_error` — matching the "bad model file" convention already used throughout the loader (`get_dsp.cpp:121,139,173`, `model_config.h:86`, and ~25 sites in `wavenet/model.cpp`). They're kept deliberately small, and in their own header rather than `NAM/util.h` so consumers don't pick up `<Eigen/Dense>` transitively.

Required-field reads are routed through them in the envelope (`get_dsp`), the WaveNet config parser, the Linear parser, and the activation config.

Values are now bounded: dimensions to `[1, 65536]`, `groups >= 1`, and array lengths to 4096. Both caps are memory-safety bounds, not claims about which models are legitimate, and are commented as such.

## Compatibility

**Any model that loads today still loads, identically.** Fields that were optional (`.value(key, default)` / `.find()`) stay optional — nothing was promoted to required — and no accepted value changes meaning. The only inputs whose behaviour changes are ones that were previously undefined behaviour or a crash.

The `\throws` documentation added to `get_dsp.h` says `std::exception` rather than `std::runtime_error`, because some malformed input still surfaces as `nlohmann::json::exception`, which derives from `std::exception` but **not** from `std::runtime_error`. Documenting the narrower type would have been wrong and could send a host into `std::terminate()`.

## Tests

`tools/test/test_model_validation.cpp` covers each rejection — missing `version`/`architecture`/`config`/`layers`, negative and oversized dimensions, empty and over-long `dilations`, `"layer1x1": {}` and `"head1x1"` missing inner keys, `"activation": {}`, `"groups_input": 0`, FiLM `groups: 0`, negative `kernel_size`, non-integral `channels`, and the exact `kMaxModelDimension` boundary — plus positive tests that a well-formed WaveNet and Linear model still load and process audio.

Verified as follows:

- Builds clean under clang 18 with `-Werror`, **Debug, Release, and `NAM_USE_INLINE_GEMM=ON`**; `run_tests` exits 0 on all three.
- `clang-format` 19 reports every changed file clean.
- **Reverting the library changes while keeping the new tests makes the suite abort at `json.hpp:22188`** — the exact `operator[]` assertion above. The tests genuinely catch the bug rather than passing incidentally.
- A deliberately injected failure confirmed all new tests actually execute (the runner registers tests by hand, so an unregistered test would silently never run).

## What this does *not* fix

Stated plainly so the scope isn't overread:

- **Truncated `weights` array is still a heap over-read.** `set_weights_` walks the vector with unchecked `*(it++)` (`conv1d.cpp:22,46,53`, `lstm.cpp:21-28`, `convnet.cpp:22-151`), and the only bounds check runs *after* everything is consumed and detects surplus, not deficit. The dimension caps here do **not** mitigate it — a 1-layer, 1-channel model with an empty `weights` array reproduces it just as easily. Fixing it properly means a bounds-checked weight reader, which re-types `set_weights_` across several headers; that's a separate PR and I'd rather not bundle it.
- **The caps bound a single field, not the product feeding an allocation.** `channels` at the 65536 cap still makes one `Conv1D` tap request ~16 GiB. This closes the UB and the trivial crashes, not memory-exhaustion DoS.
- **`convnet.cpp` and `lstm.cpp` still use unchecked `operator[]`** on required fields. Left out to keep this diff reviewable; happy to follow up.
- **Recursive `get_dsp`** (`condition_dsp` at `model.cpp`, container submodels at `container.cpp:163`) has no depth limit. Unchanged by this PR, but worth knowing about.

## Review notes

Reviewed locally by a code-review and a security pass before opening. Both initially returned *request changes* — the first version hardened the top-level fields but left the identical `JSON_ASSERT` UB live in the `layer1x1`/`head1x1` sub-objects and in `activations.cpp`, which would have made for a misleadingly-named "hardening" change. Those are fixed here, along with the `groups` divide-by-zero and the array-length bound, which the security pass surfaced.

Happy to split this further, adjust the cap values, or drop the `json_util` helper in favour of inline checks if you'd prefer a smaller surface.

### DIFF
--- a/NAM/activations.cpp
+++ b/NAM/activations.cpp
@@ -91,7 +91,19 @@ nam::activations::ActivationConfig nam::activations::ActivationConfig::from_json
   // If it's an object, parse type and parameters
   if (j.is_object())
   {
-    std::string type_str = j["type"].get<std::string>();
+    // `j` is a const reference, so `j["type"]` would resolve to nlohmann's const
+    // `operator[]`, which asserts (UB under -DNDEBUG) if "type" is missing. Look it up
+    // through `find()` instead so a missing field throws instead of crashing.
+    const auto type_it = j.find("type");
+    if (type_it == j.end())
+    {
+      throw std::runtime_error("Activation config: missing required field 'type'");
+    }
+    if (!type_it->is_string())
+    {
+      throw std::runtime_error("Activation config: field 'type' must be a string");
+    }
+    std::string type_str = type_it->get<std::string>();
     auto it = type_map.find(type_str);
     if (it == type_map.end())
     {
--- a/NAM/get_dsp.cpp
+++ b/NAM/get_dsp.cpp
@@ -9,6 +9,7 @@
 #include "registry.h"
 #include "json.hpp"
 #include "get_dsp.h"
+#include "json_util.h"
 #include "model_config.h"
 
 namespace nam
@@ -141,13 +142,17 @@ std::vector<float> GetWeights(nlohmann::json const& j)
 
 void populate_dsp_data(const nlohmann::json& config, dspData& returnedConfig)
 {
-  verify_config_version(config["version"].get<std::string>());
+  static constexpr const char* kContext = "Model file";
 
-  nlohmann::json config_json = config["config"];
+  const std::string version = nam::util::RequireValue<std::string>(config, "version", kContext);
+  verify_config_version(version);
+
+  const nlohmann::json& config_json = nam::util::RequireField(config, "config", kContext);
+  const std::string architecture = nam::util::RequireValue<std::string>(config, "architecture", kContext);
   std::vector<float> weights = GetWeights(config);
 
-  returnedConfig.version = config["version"].get<std::string>();
-  returnedConfig.architecture = config["architecture"].get<std::string>();
+  returnedConfig.version = version;
+  returnedConfig.architecture = architecture;
   returnedConfig.config = config_json;
   returnedConfig.metadata = config.value("metadata", nlohmann::json());
   returnedConfig.weights = weights;
--- a/NAM/get_dsp.h
+++ b/NAM/get_dsp.h
@@ -77,16 +77,32 @@ struct DspLoadOptions
   std::optional<bool> prewarm = std::nullopt;
 };
 
+// A note on exceptions: `.nam` files are untrusted input (they're downloaded from the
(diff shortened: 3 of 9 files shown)


sdatkinson:
Thanks! This seems useful. I added a bit of validation earlier today for #314 / motivated by https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/569.

The line diff on this is too large for me to get through at the moment but it is interesting to me. What I'd appreciate:

* Move validation to a stand-alone function instead of interleaving in the current code.
* I personally don't like "json_util.{cpp.h}" in that the validation is for `.nam` files, not JSON files (it should be treated as coincidental that `.nam` files are valid JSON). So if that can be tweaked that'd be great.

If done as I'm thinking, the diffs would be a one-liner call to your new validation functions at the starts of the `parse_config_json`s, and then the rest (which assumes happy path) requires no diffs (I think). Then the validation code can also be in one self-contained block.

If you can address these, happy to review.

sdatkinson (changes requested):
See comment

PR #319, opened 2026-08-09
Cache A2 prewarm state

### LINKED ISSUE #318: A2-fast: cache prewarm state
Prewarming A2 is currently done by processing silence. The purpose of this is to get the correct states in the convolution layers' ring buffers.

Those states could be saved for each convolution layer in an `in_channels`-length array (they should be constant in time). 

The proposed change is: When prewarming, ask the model whether it has a prewarm state computed--`bool HasCachedPrewarmState()`. If `true`, then call a `PrewarmFromCache()`; If not, run the prewarm as usual, then call some new `CacheStateAsPrewarmed()` before returning.

This should provide a substantial speedup in places where prewarming happens, like when `Reset()` is called usually.

At the top `WaveNet` class level, this amounts to the addition of a new private method; however, the objects contained within the WaveNet (layer arrays, layers, convs, head) need public methods to be accessed as the call travels down from the top.

This is something that can later be applied to other model types, but A2 (and in particular the A2 fast path) is the priority and the scope of this Issue.

A2-Lite's storage footprint is a primary concern since it fits on highly-constrained embedded systems.Thinking out loud, I believe this will add 24x3=72 more floats to the model storage for A2-Lite, which should be no problem.

sdatkinson:
## Summary
- cache one steady-state input sample for each A2 convolution and rebuild ring histories from it on later prewarms
- support both the specialized A2-fast implementation and the unconditioned generic WaveNet fallback
- preserve full silence processing for conditioned WaveNets, whose nested DSP state cannot be reconstructed from convolution caches alone
- use the correct A2-Lite and A2-Full terminology throughout the fast-path code, tests, and benchmark output

## Tests
- `cmake --build build --target run_tests -j4 && ./build/tools/run_tests`
- full debug suite with `NAM_ENABLE_A2_FAST=OFF`
- full debug suite with `NAM_A2_RING_MODE=0`
- A2-Lite and A2-Full tests first prewarm through legacy silence processing, process more than a full receptive field of audio, restore the cached prewarm state without allocations, and require exactly identical output when processing the same audio again; both A2-fast and generic WaveNet paths are covered

Resolves #318

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ endif()
 set(NAM_DEPS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/Dependencies")
 include_directories(SYSTEM "${NAM_DEPS_PATH}/eigen")
 
-# Build the specialized A2 fast-path WaveNet (A2 standard + A2 nano). When ON,
+# Build the specialized A2 fast-path WaveNet (A2-Full + A2-Lite). When ON,
 # models whose config matches the A2 shape signature are routed to a hand-optimized
 # WaveNet implementation instead of the generic one. When OFF, all models go
 # through the generic path.
--- a/NAM/conv1d.cpp
+++ b/NAM/conv1d.cpp
@@ -1,5 +1,6 @@
 #include "conv1d.h"
 #include "compiler.h"
+#include <cassert>
 #include <cstring>
 #include <stdexcept>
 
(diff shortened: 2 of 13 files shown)


sdatkinson:
Performance metrics on my M1:


Model | Old code: 100 resets | New code: 100 resets | Per reset, old → new | Speedup
-- | -- | -- | -- | --
A2-Lite | 72.981 ms | 0.859 ms | 729.81 → 8.59 µs | 85.0×
A2-Full | 517.224 ms | 1.123 ms | 5,172.24 → 11.23 µs | 460.6×

Issue #320, opened 2026-08-12
[BUG] layer1x1_post_film is ignored unless gating mode is blended


phillipmself:
wavenet::detail::Layer::Process applies layer1x1_post_film only inside the GatingMode::BLENDED branch. Under GatingMode::NONE and GatingMode::GATED the layer runs layer1x1 and never applies the FiLM.

The trainer applies it whenever layer1x1 is active, independent of gating (nam/models/wavenet/_layer_array.py):

```
layer_output = post_activation
if self._layer1x1 is not None:
    layer_output = self._layer1x1(layer_output)
    if self._layer1x1_post_film is not None:
        layer_output = self._layer1x1_post_film(layer_output, ...)
```

So a model trained with layer1x1_post_film active and non-blended gating renders differently in the runtime than it does in training. It fails silently: the config parser accepts it (it only checks that layer1x1_post_film.active implies layer1x1.active), the FiLM is constructed, and set_weights_ still consumes its weights — so the rest of the blob is read correctly and nothing is misaligned. The modulation is simply skipped.

test_layer1x1_post_film_active() nominally covers this case (it uses GatingMode::NONE) but asserts only that the output is finite, and its FiLM weights make the modulation an identity, so it cannot catch it.

Found while building end-to-end parity fixtures against the trainer with every FiLM site active under GatingMode::NONE. PR follows.

PR #321, opened 2026-08-12
Apply layer1x1_post_film under every gating mode

phillipmself:
Fixes the divergence described in #320.

layer1x1_post_film was applied only inside the BLENDED branch of Layer::Process. Under GatingMode::NONE and GatingMode::GATED the modulation was silently dropped, while the trainer applies it whenever layer1x1 is active regardless of gating.

Nothing else changes: the FiLM was already constructed and its weights already consumed by set_weights_, so weight-blob alignment is unaffected. This only adds the missing application in the two branches that lacked it.

Test: test_layer1x1_post_film_active() used GatingMode::NONE but asserted only finiteness, with FiLM weights that made the modulation an identity — so it passed either way. Added test_layer1x1_post_film_is_applied_for_every_gating_mode, which builds an otherwise-identity layer, runs it with FiLM scale 1.0 and 2.0, and asserts the output moves — for NONE, GATED and BLENDED. Verified it fails on main and passes with this change.

Compatibility note: this changes rendered output for any existing model that has layer1x1_post_film active with non-blended gating. Those models were rendering incorrectly relative to their training; this brings them in line.

### DIFF
--- a/NAM/wavenet/model.cpp
+++ b/NAM/wavenet/model.cpp
@@ -241,6 +241,11 @@ void nam::wavenet::detail::Layer::Process(const Eigen::MatrixXf& input, const Ei
     if (this->_layer1x1)
     {
       this->_layer1x1->process_(this->_z, num_frames);
+      if (this->_layer1x1_post_film)
+      {
+        Eigen::MatrixXf& layer1x1_output = this->_layer1x1->GetOutput();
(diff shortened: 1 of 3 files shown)


sdatkinson:
Thanks for catching this! I believe that the change here is valid, but I like the non-duplication in #322. But I also like the test here.

So I'm going to merge 321, then change this PR to only include the added test, then merge it as well.

phillipmself:
Thanks!

PR #322, opened 2026-08-20
Fix layer1x1 post FiLM for non-blended gating

### LINKED ISSUE #320: [BUG] layer1x1_post_film is ignored unless gating mode is blended

Helshr:
Fixes #320

## Summary

`layer1x1_post_film` was only applied for `GatingMode::BLENDED`. This caused models using the FiLM with `GatingMode::NONE` or `GatingMode::GATED` to produce different output from the trainer.

The FiLM is now applied whenever `layer1x1` is active, regardless of the gating mode.

## Tests

- Added non-identity regression coverage for `GatingMode::NONE`
- Added regression coverage for `GatingMode::GATED`
- `./build/tools/run_tests` passes successfully

### DIFF
--- a/NAM/wavenet/model.cpp
+++ b/NAM/wavenet/model.cpp
@@ -279,12 +279,15 @@ void nam::wavenet::detail::Layer::Process(const Eigen::MatrixXf& input, const Ei
     if (this->_layer1x1)
     {
       this->_layer1x1->process_(this->_z.topRows(bottleneck), num_frames);
-      if (this->_layer1x1_post_film)
-      {
-        Eigen::MatrixXf& layer1x1_output = this->_layer1x1->GetOutput();
-        this->_layer1x1_post_film->Process_(layer1x1_output, condition, num_frames);
-      }
     }
+  }
+
+  // layer1x1_post_film is independent of the gating mode and must be applied
+  // whenever layer1x1 is active.
+  if (this->_layer1x1 && this->_layer1x1_post_film)
+  {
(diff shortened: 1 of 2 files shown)


sdatkinson:
Thanks! I like the logic behind your change to the source, and I like the added tests in #321. In fact I believe there's a minor bug in your test. I'll fix that, merge this, then square it up with #321 and merge the test from that one.

sdatkinson (approved):
Thanks!

